PR #7, opened 2020-09-22
update biotransformer versioning to iuc standard

martenson:
add missing requirement
drop extra whitespace


trachtok (approved):

martenson:
@trachtok Splendid. All we are missing now is a test.

PR #12, opened 2020-09-24
Cleaned xml wrapper and added test.

trachtok:
I have cleaned the xml wrapper and removed input/output formats that are not used (and not working with current Python script). Also test was added to the xml wrapper (test data included).


tools/biotransformer/biotransformer.xml
```diff
@@ -51,43 +38,46 @@
     <param name="tolerance" type="float" value="0.01" label="Mass tolerance for metabolite identification."/>
 
     <conditional name="input">
-    	<param name="type_selector" type="select" label="Input format [SMILES, SDF, CSV].">
-        	<option value="smile" selected="true">SMILES</option>
-		<option value="sdf">SDF</option>
-		<option value="csv">CSV</option>
+    	<param name="type_selector" type="select" label="Input format [CSV].">
+		<option value="csv" selected="true">CSV</option>
```
martenson, line 42: Given there is no other option to choose from we can drop the whole selector I think.

martenson:
at the moment this seems to be failing in pandas because it is trying to parse an empty file

```
tmp2 = pandas.read_csv(out.name)
....
  File "pandas/_libs/parsers.pyx", line 540, in pandas._libs.parsers.TextReader.__cinit__
pandas.errors.EmptyDataError: No columns to parse from file
```

tools/biotransformer/biotransformer.xml
trachtok: I left it there because in future I plan to add input type of simple SMILES string, but it is not a priority now.

martenson: For the moment this is a dead code and I think we should remove it. You can always go back to this version using git when you decide to expand the implementation.

tools/biotransformer/biotransformer.xml
```diff
@@ -51,43 +38,46 @@
...
-    	<when value="smile">
-        	<param name="input_format" argument="-ismi" type="text" label="SMILES (example string format: CC(C)C1=CC=C(C)C=C1O )."/>
-    	</when>
 	<when value="csv">
-		<param name="input_format" argument="-csv" type="data" format="csv" label="CSV with one molecule (SMILES format) per line."/>
+		<param name="input_format" argument="-icsv" type="data" format="csv" label="CSV with one molecule (SMILES format) per line."/>
 	</when>
     </conditional>
 
+    <!-- If we would like to allow user to choose between format - not available now
```
martenson, line 49: Same here, please drop the commented out code, you can go back for it any time in the future using git.

tools/biotransformer/biotransformer.xml
```diff
@@ -11,25 +11,12 @@
  -s $steps
  -t $tolerance
  -a
- #if $input.type_selector == "sdf"
-	-isdf $input.input_format
- #else if $input.type_selector == "csv"
+ #if $input.type_selector == "csv"
 	-icsv $input.input_format
```
martenson, line 15: Using `$input_name` would be more clear.

martenson (commented):
the tool's version needs a bump to `1.1.5+galaxy1`

martenson:
One of biotransformer dependencies has an issue with the platform on which our CI runs.
```
Error loading JNI InChI native code.
You may need to compile the native code for your platform.
See http://jni-inchi.sourceforge.net for instructions.
```
More details: http://jni-inchi.sourceforge.net/faq.html

We will merge this even with failing test and see how it behaves on other platforms.

PR #17, opened 2020-10-01
Change test results comparison condition.

smartx-usman:
This PR will also help to test newly pushed docker image (Docker hub) for C stack error during CI process.

### DIFF
--- a/tools/retip/retip_apply.xml
+++ b/tools/retip/retip_apply.xml
@@ -22,7 +22,7 @@
             <param name="descr_train" value="descriptors.feather"/>
             <param name="model_hdf5" value="model.hdf5"/>
             <param name="input_smiles" value="input.tsv"/>
-            <output name="output1" file="output.tsv" ftype="tabular"/>
+            <output name="output1" file="output.tsv" ftype="tabular" compare="sim_size" delta="1000"/>
         </test>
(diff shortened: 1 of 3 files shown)


tools/retip/retip_apply.xml
```diff
@@ -22,7 +22,7 @@
             <param name="descr_train" value="descriptors.feather"/>
             <param name="model_hdf5" value="model.hdf5"/>
             <param name="input_smiles" value="input.tsv"/>
-            <output name="output1" file="output.tsv" ftype="tabular"/>
+            <output name="output1" file="output.tsv" ftype="tabular" compare="sim_size" delta="1000"/>
```
martenson, line 25: This file has 602 Bytes, comparing with a delta of 1000 Bytes will always pass. What are the possible differences in this tabular file that are happening on a run basis?

smartx-usman: I am getting almost the same size on every run (e.g., every time it is approx. 602 bytes). Would any number between 10 ~ 50 bytes be a reasonable delta?

smartx-usman: This test results in just 5 lines. But in each test, each line changes because some numerical values are predicted and have small differences.

PR #20, opened 2020-10-05
aplcms:6.6.6-galaxy1 - add a basic tests & fix HDF format support

martenson:
and bump container


xtracko:
The test looks mostly fine. The data are not great, but not terrible. I don't fancy comparing the h5 files for file size similarity. Is there any way to invoke a script?
I have made a slight change to explicitly fix the order of the test files.
I'm not sure If this build will succeed because of the known problem with the h5 files.

martenson:
@xtracko I don't like the size comparison either. There are some assertions we can maybe use (like `has_h5_attribute` or `has_h5_keys `): https://docs.galaxyproject.org/en/master/dev/schema.html#tool-tests-test-output-assert-contents

martenson:
Also, how would a test for the hybrid version look like?

xtracko:
> 
> 
> Also, how would a test for the hybrid version look like?

It would be similar, we only need one more file which I need to easily recreate. We have lost it (like most of the files) when we migrated the galaxy to IM infrastructure.

I am also hoping that apLCMS is not build upon random generators (I dont know the whole codebase). Fortunately, I've tested the unsupervised method on 3 different machines and it gave me similar results.

martenson:
>  it gave me similar results.

Similarity is fairly hard to test with tools we have at our disposal. If we don't want to do the size comparison we can count rows and check for specific h5 attribute values and keys. But from what I understand the crucial values will always be a little different run from run.

PR #23, opened 2020-10-08
Support both tsv and h5 data input types and add new wrapper

smartx-usman:


martenson (approved):
lgtm, thank you @smartx-usman !

PR #25, opened 2020-10-12
Add new input field for filtering mode.

smartx-usman:


martenson:
Thanks @smartx-usman !

PR #36, opened 2020-11-06
add wrapper for RAMClustR

hechth:
Started wrapper for RAMClustR

* Implemented docker container for tool
* Forwarded basic parameters to script
* Handles XCMS object as input
* Outputs MSP files and CSV with abundances


tools/ramclustr/ramclustr.xml
```diff
@@ -0,0 +1,62 @@
+<tool id="ramclustr" name="RAMClustR" version="@TOOL_VERSION@+galaxy1">
+    <macros>
+        <import>ramclustr_macros.xml</import>
+    </macros>
```
xtracko, line 4: If you have only one tool variant, there is no need to have the parameters in the macros file. Generally, the macros file should be used for code reuse. If you plan to have multiple RAMClustR tools, ignore this comment.

You can eighter substitute the @TOOL_VERSION@ token with an actual value or you can just place the <token> tag in <macros> block inside this file. I would do the latter.

tools/ramclustr/ramclustr.xml
```diff
@@ -0,0 +1,62 @@
...
+
+    <inputs>
+        <param name="input_xcms" label="input_xcms" type="data" format="RData" help=": containing grouped feature data for clustering by ramclustR" />
+        <expand macro="parameters" />
+    </inputs>
+
+    <command detect_errors="aggressive"><![CDATA[
+    
+        Rscript 
+            -e 'obj <- load("$input_xcms")'
```
xtracko, line 18: The purpose of a Galaxy tool is to make it usable by the non-IT person and to integrate it into the Galaxy ecosystem. Hence, exposing the R-specific format to the wrapper's interface is not suitable. If the input/output is tabular then expose it as a tsv (preferable) or csv files. If the tabular format contains a lot of high precision decimal numbers choose some binary tabular format (for instance h5). If the input is of some unknown structure then we have a problem and we should have a discussion on this.

tools/ramclustr/ramclustr.xml
```diff
@@ -0,0 +1,62 @@
...
+                qc.inj.range = $params.qc_inj_range,
+                normalize = "$params.normalize",
+                minModuleSize = $params.minModuleSize,
+                linkage = $params.linkage,
+                mzdec = $params.mzdec,
+                cor.method = "$params.cor_method",
+                rt.only.low.n = $params.rt_only_low_n,
+                replace.zeros = $params.replace_zeros
+                )'
+            -e 'save(x, file = "$result")'
```
xtracko, line 51: Same comment as on the line 18.

tools/ramclustr/ramclustr.xml
```diff
@@ -0,0 +1,62 @@
...
+    ]]>
+    </command>
+
+    <outputs>
+        <data name="result" format="RData" />
+    </outputs>
+
```
xtracko, line 58:
```suggestion
<tests>
  <test>
    <param ... />
  </test>
</tests>
```

Please add some tests - to retrieve the reference outputs just run it once on your machine on some open-sourced dataset. One such mzml dataset is in the `aplcms` directory.

tools/ramclustr/ramclustr_macros.xml
```diff
@@ -0,0 +1,47 @@
...
+            <param label="rt_only_low_n" name="rt_only_low_n" type="boolean" truevalue="TRUE" falsevalue="FALSE" value="TRUE" help="At low injection numbers, correlational relationships of peak intensities may be unreliable.  by defualt ramclustR will simply ignore the correlational r value and cluster on retention time alone.  if you wish to use correlation with at n less than 5, set this value to FALSE." />
+            <param label="fftempdir" name="fftempdir" type="text" optional="true" help="valid path: if there are file size limitations on the default ff pacakge temp directory  - getOptions('fftempdir') - you can change the directory used as the fftempdir with this option." />
+            <param label="replace_zeros" name="replace_zeros" type="boolean" truevalue="TRUE" falsevalue="FALSE" value="TRUE" help="NA, NaN, and Inf values are replaced with zero, and zero values are sometimes returned from peak peaking.  When TRUE, zero values will be replaced with a small amount of noise, with noise level set based on the detected signal intensities for that feature. " />
+        </section>
+    </xml>
```
xtracko, line 46: I haven't checked these params one by one, but they seem fine. Moreover, they are documented properly :)

xtracko (commented):
I am not an expert on tool wrapping (I wrapped only 3 tools) so please don't take my comments literally.

That being said I would like to pinpoint one thing. The input/output files. Having the requirement of accepting only the `rdata` format is unacceptable (Yes, I know that Workflow4Metabolomic are doing that). For development use, you can have the `rdata` format but for production use, we need to choose something proper (TSV or H5 or something not R specific with wide user support).

I hope you'll find my comments helpfull ;)

tools/ramclustr/ramclustr.xml
hechth: There might be different tools or the current tool will be split into different steps, so the different tools might need different combinations of parameters.

tools/ramclustr/ramclustr.xml
hechth: The tool itself takes an XCMS R object as input, that wasn't my choice. It can be directly wired in from XCMS in Galaxy.

tools/ramclustr/ramclustr.xml
hechth: The output of the tool is still handled preliminary.

tools/ramclustr/ramclustr.xml
hechth: There is yet no way to run the tool completely, I'd like to include it in our version of Galaxy for further testing.

martenson: @hechth How does a tool that cannot be run yet benefit from being available on UMSA Galaxy?

tools/ramclustr/ramclustr.xml
xtracko: Ok, therefore, I suggest having multiple input/output types. You can specify mulltiple formats by `<param type="data" format="rdata,tsv">` and handle these formats differently in the <command> segment. If the input is only a table (or several tables) it's worth the shot.

tools/ramclustr/ramclustr.xml
hechth: I wired in the parameters and can't yet tell if everything runs smoothly as the parameters have to be tuned for the data, which is something where I'd like @ElliottJP to have a look at.

If no non completely tested tool should be in the Galaxy I'll test these things differently.

martenson: I am fine with making this available on UMSA, we are in a development phase and everything there should be considered beta at best. However, I would also like to see a test if possible, since that documents how to use the tool and makes it much more accessible and easier to reason about I think. 

If you are saying test is not helpful at the moment, that is fine with me.

tools/ramclustr/ramclustr_xcms.xml
```diff
@@ -0,0 +1,95 @@
...
+            -e 'RAMClustR::write.msp(x, one.file = $required.merge_msp)'
+            -e 'write.csv(x\$SpecAbund, file="$spec_abundance", row.names=TRUE)'
+    ]]>
+    </command>
+
+    <outputs>
+        <data label="${tool.name} on ${on_string}" name="result" format="RData" />
+        <data label="Spec Abundance of ${on_string}" name="spec_abundance" format="csv" />
+        <data label="Metadata for ${tool.name} on ${on_string}" name="method_metadata" format="txt"/>
+        <expand macro="output_msp"/>
```
xtracko, line 59: Consider using `hidden` attribute (see https://docs.galaxyproject.org/en/master/dev/schema.html#id65) for non-important outputs (it will be only hidden from the user in the UI) (Galaxy doesn't group outputs by the tool so the Galaxy history might get messy).

xtracko (commented):
Just a minor comments. It depends on you whether to bother with them.

tools/ramclustr/ramclustr_xcms.xml
```diff
@@ -0,0 +1,95 @@
...
+    <macros>
+        <import>ramclustr_macros.xml</import>
+    </macros>
+
+    <requirements>
+        <container type="docker">recetox/ramclustr:1.1.0-recetox0</container>
+    </requirements>
+
+    <inputs>
+        <param name="input_xcms" label="input_xcms" type="data" format="RData" help=": containing grouped feature data for clustering by ramclustR" />
```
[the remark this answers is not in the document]
hechth: I changed it to this subtype since it requires this specific object

martenson:
this PR is superseded by https://github.com/RECETOX/galaxytools/pull/54

PR #54, opened 2020-12-02
wrap RAMClustR

martenson:
This is the same PR as https://github.com/RECETOX/galaxytools/pull/36 but with cleaned commit history. Github got confused by some merges/rebases in the previous one.

Please review the comments in the original PR before merging.


martenson:
I reviewed the point of @xtracko from the original PR, afaik all are either resolved or marked as temporary solutions.

I tried swapping the requirement for [conda](https://bioconda.github.io/recipes/r-ramclustr/README.html) package version 1.0.9, but it does not work for now. The version difference may have something to do with it.

PR #63, opened 2020-12-08
WIP: execute SQL query on binary tabular files

xtracko:
Features:
- [x] convert between various text and binary tabular datatypes
- [x] execute an SQL SELECT
- [x] execute SQL JOIN on many input files of different format
- [x] choose an output format


tools/query/.shed.yml
```diff
@@ -0,0 +1,6 @@
+name: query
```
martenson, line 1: the name is very generic, this will be super difficult to search for

tools/query/query.xml
```diff
@@ -0,0 +1,134 @@
...
+        </xml>
+        <xml name="internal_name" token_format="">
+            <when value="@FORMAT@">
+                <param name="name" type="text" help="Specify the format's internal name of the table."/>
+            </when>
+        </xml>
+    </macros>
+
+    <requirements>
+        <requirement type="package">click</requirement>
```
martenson, line 20: the requirements should be versioned for many reasons including reproducibility

tools/query/query.xml
```diff
@@ -0,0 +1,134 @@
...
+            </param>
+            <param name="name" type="text" optional="true">
+                <label>Table name</label>
+                <help>Choose an unique name for the table which then you may use in the SQL Query. By default tables will be named as: t0, t1, ...</help>
+            </param>
+        </repeat>
+        <param name="query" type="text" area="true">
+            <label>SQL Query</label>
+            <help>
+                Type an arbitrary SQL SELECT to perform on the input tables. The input tables may be referred by their given names.
```
martenson, line 89: arbitrary SQL query could be scary, we need to check for security implications of giving such tool to users

tools/query/query.xml
```diff
@@ -0,0 +1,134 @@
...
+            #import json
+            #echo $json.dumps($config)
+        ]]></configfile>
+    </configfiles>
+
+    <inputs>
+        <repeat name="tables" title="Table">
+            <param name="file" type="data" format="csv,tsv,h5,feather,parquet,sqlite" label="File"/>
+            <param name="format" type="select" optional="true">
+                <label>Force the file format</label>
```
martenson, line 77: I assume this is here only because of the parquet format and will be removed. Is that correct?

tools/query/query.xml
bernt-matthias: As long as `pandasql` can not be used to execute code it should be fine since the input is opened read only. The worst thing that I could imagine is that the output is messed up.

tools/query/query.xml
```diff
@@ -0,0 +1,134 @@
...
+            <option value="csv">csv</option>
+            <option value="tsv">tsv</option>
+            <option value="h5">h5</option>
+            <option value="feather">feather</option>
+            <option value="parquet">parquet</option>
+            <option value="sqlite">sqlite</option>
+        </xml>
+        <xml name="internal_name" token_format="">
+            <when value="@FORMAT@">
+                <param name="name" type="text" help="Specify the format's internal name of the table."/>
```
bgruening, line 14: a sanitizer would be nice to add I think, as in every text field.

tools/query/query.xml
bgruening: The code looks actually small enough to inline the entire script as a configfile. This would reduce the dependencies and also the code in general as we do not need to pass parameters.

tools/query/query.xml
```diff
@@ -0,0 +1,134 @@
...
+                'query': str($query),
+                'result': {
+                    'path': str($output),
+                    'name': str($result.name) if str($result.format) in ('h5', 'hdf', 'sqlite') and $result.name else None,
+                    'format': str($result.format)
+                }
+            }
+
+            #import json
+            #echo $json.dumps($config)
```
bgruening, line 69: Why not including here (or in another config file) the python script and generate it on the fly?

tools/query/query.xml
bgruening: Its also worth looking at https://github.com/galaxyproject/tools-iuc/blob/master/tools/gemini/gemini_macros.xml#L101 and Gemini in general as those tools also deal with sqlite.

tools/query/query.xml
```diff
@@ -0,0 +1,134 @@
...
+        </conditional>
+    </inputs>
+
+    <outputs>
+        <data name="output">
+            <change_format>
+                <when input="result.format" value="h5" format="h5"/>
+                <when input="result.format" value="csv" format="csv"/>
+                <when input="result.format" value="tsv" format="tsv"/>
+                <when input="result.format" value="feather" format="feather"/>
```
bgruening, line 110: feather and parquet needs to be added to Galaxy isn't it?

martenson: @bgruening Parquet for sure, feather we might not even want.

tools/query/.shed.yml
```diff
@@ -0,0 +1,6 @@
+name: query
+owner: recetox
+remote_repository_url: "https://github.com/RECETOX/galaxytools/tree/master/tools/query"
+categories:
+  - Text Manipulation
+description: "Execute an SQL statement on a set of tables"
```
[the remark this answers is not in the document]
martenson: What viz would this be?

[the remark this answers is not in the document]
bgruening: https://github.com/galaxyproteomics/tools-galaxyp/tree/master/visualizations/sqlviewer

tools/query/.shed.yml
xtracko: Since we have started the more serious discussion with all of you involved, I propose to contribute this working prototype to a similar existing IUC tool [Query Tabular](https://github.com/galaxyproject/tools-iuc/blob/master/tools/query_tabular/query_tabular.xml) which works only on text tabular files. Otherwise, I agree that the name is too generic.

tools/query/query.xml
xtracko: Yup, that's right.

tools/query/.shed.yml
bgruening: Uh, I :heart:  that!

tools/query/query.xml
xtracko: Parquet and Feather are similar formats (Feather is only faster) - if we add Parquet we should also add Feather (at least for the Galaxy community). Both of them come from the same vendor and both of them are supported by the official python library pyarrow. The additional work is minimal.

blankenberg:
Just wanted to also briefly mention query tabular (I don't think it handles hdf5, etc): https://f1000research.com/articles/7-1604

PR #67, opened 2020-12-13
Add tool to convert HDF to csv

maximskorik:
Tool to convert hdf output of apLCMS to csv input for RamClustR. See issue #51 for more info. 
* planemo test --docker OK
* planemo serve --docker OK
* planemo lint > Failed linting - INFO: File validates against XML schema. No clue how to fix that yet.

Table is of format as described in original RamClustR [repo](https://github.com/cbroeckl/RAMClustR/blob/master/vignettes/RAMClustR.Rmd), i.e. sample names in the first column. Column names contain the mz value and the retention time separated by an underscore "_".

image

PR #70, opened 2020-12-15
HDF to csv converter. Change tool dependencies to be resolved by conda

maximskorik:
@martenson, could you take a look please. Does one have to specify conda collection when setting up dependencies (e.g. "conda-forge") in xml, or it is fine the way it is?

followup to https://github.com/RECETOX/galaxytools/pull/67

### DIFF
--- a/tools/hdf_converter/hdf_converter.xml
+++ b/tools/hdf_converter/hdf_converter.xml
@@ -1,7 +1,9 @@
-<tool id="apCLMS_to_RamClustR_converter" name="apLCMS to RamClustR converter" version="0.0.1" python_template_version="3.5">
+<tool id="aplcms_to_ramclustr_converter" name="apLCMS to RamClustR converter" version="0.0.1" python_template_version="3.5">
     <description>converts aplcms HDF output to RamClustR csv input</description>
     <requirements>
-        <container type="docker">maxskorik/hdf_converter</container>
+        <requirement type="package" version="3.7">python</requirement>
+        <requirement type="package" version="1.1.5">pandas</requirement>
+        <requirement type="package" version="3.6.1">pytables</requirement>
     </requirements>
     <command detect_errors="exit_code"><![CDATA[
         python $__tool_directory__/hdf_converter.py --dataframe '${hdf_dataframe}' --table '${table_name}' '${out_file}'
@@ -24,6 +26,12 @@
         </test>
     </tests>
     <help><![CDATA[
-        Tool to convert apLCMS output to csv file formated in accordance to RamClustR input requirements.
+        Tool to convert apLCMS output to csv file formated in accordance to `RamClustR (CSV) <https://umsa.cerit-sc.cz/root?tool_id=testtoolshed.g2.bx.psu.edu/repos/recetox/ramclustr_csv/ramclustr_csv/1.1.0+galaxy0>`_ input requirements.
+
+        **Workflow position**
+
+        - Upstream tool: `apLCMS <https://github.com/RECETOX/galaxytools/tree/master/tools/aplcms>`_  Unsupervised or Hybrid
+
+        - Downstream tool: `RamClustR (CSV) <https://github.com/RECETOX/galaxytools/tree/master/tools/ramclustr>`_
     ]]></help>
 </tool>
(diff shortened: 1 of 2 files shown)


maximskorik:
I accidentally pushed irrelevant files from my local directory. Sorry for the mess.

tools/hdf_converter/hdf_converter.xml
```diff
@@ -1,7 +1,8 @@
 <tool id="apCLMS_to_RamClustR_converter" name="apLCMS to RamClustR converter" version="0.0.1" python_template_version="3.5">
     <description>converts aplcms HDF output to RamClustR csv input</description>
```
martenson, line 2: Is this true every time or is this applicable only to our wrappers of said tools? Maybe we could link them in `help` section to clarify this.

tools/hdf_converter/hdf_converter.xml
```diff
@@ -1,7 +1,8 @@
 <tool id="apCLMS_to_RamClustR_converter" name="apLCMS to RamClustR converter" version="0.0.1" python_template_version="3.5">
```
martenson, line 1: the `id` here should match the one in .shed.yml

tools/hdf_converter/hdf_converter.xml
```diff
@@ -1,7 +1,8 @@
...
     <description>converts aplcms HDF output to RamClustR csv input</description>
     <requirements>
-        <container type="docker">maxskorik/hdf_converter</container>
+        <requirement type="package" version="1.1.5">pandas</requirement>
+        <requirement type="package" version="3.6.1">pytables</requirement>
```
martenson, line 5: I'd explicitly declare Python here, you probably get one with pandas/pytables but this way you have no control over version.

hechth: How do you specifically control that is is a python package?

martenson: @hechth I don't understand the question, what I am suggesting is to include something like: `<requirement type="package" version="3.7">python</requirement>`

tools/hdf_converter/hdf_converter.xml
```diff
@@ -24,6 +26,12 @@
         </test>
     </tests>
     <help><![CDATA[
-        Tool to convert apLCMS output to csv file formated in accordance to RamClustR input requirements.
+        Tool to convert apLCMS output to csv file formated in accordance to `RamClustR (CSV) <https://umsa.cerit-sc.cz/root?tool_id=testtoolshed.g2.bx.psu.edu/repos/recetox/ramclustr_csv/ramclustr_csv/1.1.0+galaxy0>`_ input requirements.
+
+        **Workflow position**
+
+        - Upstream tool: `apLCMS-unsupervised <https://umsa.cerit-sc.cz/root?tool_id=testtoolshed.g2.bx.psu.edu/repos/recetox/recetox_aplcms_unsupervised/recetox_aplcms_unsupervised/6.6.6+galaxy1>`_ or `apLCMS-Hybrid <https://umsa.cerit-sc.cz/root?tool_id=testtoolshed.g2.bx.psu.edu/repos/recetox/recetox_aplcms_hybrid/recetox_aplcms_hybrid/6.6.6+galaxy1>`_
```
martenson, line 33: These links are going to change as early as we switch to using the Main Tool Shed (and in many other circumstances). I recommend using links to github development repository as the canonical reference.

martenson: so e.g. https://github.com/RECETOX/galaxytools/tree/master/tools/ramclustr

martenson:
> Does one have to specify conda collection when setting up dependencies (e.g. "conda-forge") in xml, or it is fine the way it is?

Galaxy uses [this](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/galaxy.yml.sample#L321) order of conda channels by default. So if the package is in `conda-forge` it will be installed from there. The order is by design.

maximskorik:
> > Does one have to specify conda collection when setting up dependencies (e.g. "conda-forge") in xml, or it is fine the way it is?
> 
> Galaxy uses [this](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/galaxy.yml.sample#L321) order of conda channels by default. So if the package is in `conda-forge` it will be installed from there. The order is by design.

Alright, thanks for info

hechth (approved):

martenson (approved):
I took the liberty to squash this into a single commit, once the tests pass we can merge.

Issue #72, opened 2020-12-16
apLCMS to RAMClustR: Tool fails on actual apLCMS output


hechth:
The tool failst if I try to get the "Aligned Peaks" table or "Peaks" table, see the attached log files.
[2020_12_16_apLCMS_RAMClustR_error_log_aligned_peaks.txt](https://github.com/RECETOX/galaxytools/files/5701368/2020_12_16_apLCMS_RAMClustR_error_log_aligned_peaks.txt)
[2020_12_16_apLCMS_RAMClustR_error_log_peaks.txt](https://github.com/RECETOX/galaxytools/files/5701369/2020_12_16_apLCMS_RAMClustR_error_log_peaks.txt)

@martenson I used an input imported from another history - might that cause this problem?

Update 2020/12/16/10:11
The error was related to an invalid history item.
I replaced it with a valid h5 file, but the tool fails without error message.

martenson:
@hechth That looks like a Galaxy error, the line `    raise FileNotFoundError(f"File {path_or_buf} does not exist")` is very suspicious.

Could you provide me with the dataset or share the history?

hechth:
@martenson Here is the history: https://umsa.cerit-sc.cz/u/6a49c27ceeb84beea728267191c7a8e8/h/aplcms---ramclustr

maximskorik:
I thought there was an issue with the wrapper or the fact that Pandas cannot read '.dat' files (it does as it turned out). I executed the tool in the terminal and when I tried to inspect the csv I got a message from csv-reader stating that it may fail handling that big number of columns. So I thought that may have something to do with failing the execution. I then tried to simply upload the csv output of the converter to Planemo and our Galaxy - both failed.
@martenson is there a chance that galaxy has some sort of limitation on csv size?

maximskorik:
However, if I don't specify the file format, when uploading it uploads

maximskorik:
For example, when executed on data that Helge's sent me it made csv with 131629 columns

maximskorik:
@hechth, I created a PR #74 for the issue. Seems like I was concatenating wrong columns.
 
I was able to run RamClutsR in a container on an output csv as follows:
`> experiment <- defineExperiment(force.skip=TRUE)
> RC1 <- ramclustR(ms = paste0(getwd(), "/apLCMS-test.csv"), featdelim = "_", st = 5, ExpDes=experiment, sampNameCol = 1)`

It returned:
`organizing dataset 
[...]

PR #74, opened 2020-12-21
Fix "aplcms -> ramclutsr converter" to work with real data

maximskorik:
Fix for issue #72 
* change tool to concatenate 'mz' and 'rt' columns (concatenated 'mz' and 'time' for each sample before)
* change test data to be representative of real data
* use argparse instead of optparse
* refactor script


tools/hdf_converter/hdf_converter.py
```diff
@@ -9,55 +9,49 @@
...
 
-parser = optparse.OptionParser()
-parser.add_option("--dataframe", help="Name of hdf dataframe")
-parser.add_option("--table", help="Name of a table in the dataframe")
-(options, args) = parser.parse_args()
+parser = argparse.ArgumentParser()
+parser.add_argument("--dataframe", help="Name of hdf dataframe")
+parser.add_argument("--table", help="Name of a table in the dataframe")
+parser.add_argument('output')
+args = parser.parse_args()
```
hechth, line 16: I think this part maybe has to move into main in order to be evaluated properly. This maybe causes Galaxy to fail running the tool?

hechth (changes requested):
Try parsing the args in main and also passing the args to main, I think this is the problem here.

tools/hdf_converter/hdf_converter.py
maximskorik: As we discussed earlier I don't think these changes are necessary since the tool works as it is. I think we can leave it in this state for now.

martenson:
Could we make the 17MB test file way smaller?  Under 1MB ideally, the repository becomes unwieldy if it grows too much with test data.

martenson:
@maximskorik thanks for replacing the test data, this PR now needs to be squashed when merging, in order to remove the commits that contain the dropped file.

maximskorik:
> @maximskorik thanks for replacing the test data, this PR now needs to be squashed when merging, in order to remove the commits that contain the dropped file.

Somehow instead of squashing existing commits I've created a new one. Finding out how to fix that.

martenson:
You can squash on merge using github UI.

Fix__aplcms_-__ramclutsr_converter__to_work_with_real_data_by_maximskorik_·_Pull_Request__74_·_RECETOX_galaxytools


On cli you'd do something like `git rebase -i master` and then put `squash` in front of commits you want to squash

martenson:
But beware that squashing rewrites history, so if you are collaborating on a branch you'll need to force push and disconnect from others. It is the thing you want to do the last (after tests, after all reviews) before you merge.

maximskorik:
> But beware that squashing rewrites history, so if you are collaborating on a branch you'll need to force push and thus disconnect from others. It is the thing you want to do the last (after tests, after all reviews) before you merge.

Thank you for the advice!

martenson:
The first CI run after merging this failed so it did not get to the TTS, I re-started it and it succeeded. The TTS version is now up to date.

PR #75, opened 2021-01-06
added new tool filter_compounds

trachtok:
I have created a new simple tool for filtering out organometallics (compounds with atoms other than C|N|O|P|F|S|I|B|Si|Se|Cl|Br|Li|Na|H|K) or anorganic (compounds without C) compounds. Python script takes as an input a) list of SMILES or b) indexed table of SMILES (so that it is compatible with our QSAR-ready SMILES pipeline), translates SMILES to InChI (using OpenBabel) and uses InChI string for a simple regex search. Using InChI showed to be a better option than scanning SMILES directly, because InChI contains at the specific position in its string a simple chemical formula (summary of atoms). Scanning directly SMILES showed to be a problem as it contains specific characters and lower/upper letters to describe 3D structure, which complicated the regex search significantly.


tools/filter_compounds/filter_compounds.xml
```diff
@@ -0,0 +1,52 @@
+<tool id="filter_orgmet_anorg" name="Filter organometallics and/or anorganics" version="@TOOL_VERSION@+galaxy@GALAXY_VERSION@">
+    <description>from a library of compounds</description>
+    <!--parallelism method="multi" split_inputs="infile" split_mode="to_size" split_size="50000" shared_inputs="" merge_outputs="outfile"></parallelism-->
+    <macros>
+        <import>macros.xml</import>
+        <token name="@GALAXY_VERSION@">1</token>
```
martenson, line 6: The common practice is to start with `0` for initial release of Galaxy tool. Please consider adopting it.

tools/filter_compounds/filter_compounds.py
```diff
@@ -0,0 +1,61 @@
+import re
+import argparse
+
+from openbabel import openbabel, pybel
+openbabel.obErrorLog.StopLogging()
```
martenson, line 5: This seems suspicious, why do we need to stop logging here?

martenson (commented):
We need a `.shed.yml` file in order to test and upload& maintain this in the Galaxy Tool Shed.

tools/filter_compounds/filter_compounds.xml
trachtok: I will changed that, sorry my bad, did not notice it starts from 0.

tools/filter_compounds/filter_compounds.py
trachtok: I adopted this approach from the chemicaltoolbox tools (all that use OpenBabel contain this line), bacause otherwise Galaxy crashes with error due to logs produced by OpenBabel (that were not actually errors).

martenson: I doubt Galaxy crashes because of this. Do you mean the job will result in error because openbabel writes non-errors to stderr?

trachtok: Yes, I meant the job. I tried to remove the lines in my testing galaxy environment and the job went through, which is strange because I remember quite clearly that before Christmas the job failed and adding the line suppressing logging helped (also, some warnings about compound structure that OpenBabel quite commonly shows is now part of the stderr output, but the job is nevertheless successful). I tested also the Remove counterions and fragments tool I modified before that is part of Chemicaltoolbox, removed the `.StolLogging()` and it works as well. So I am not really sure what happed here...

martenson (approved):

PR #77, opened 2021-01-13
Fixed RAMClustR CSV by setting idmsms to NULL if not set

### LINKED ISSUE #76: Change idmsms to be handled as optional
https://github.com/RECETOX/galaxytools/blob/ad6ca51c5fee52d533cfd6b8ccef5e44027ed298/tools/ramclustr/ramclustr_csv.xml#L21

This causes a bug in the CSV based version of RAMClustR.

hechth:
Fixed RAMClustR CSV by setting idmsms to NULL if not set, because the script tries to load the file even if it is 'None'.

fixes #76

### DIFF
--- a/tools/ramclustr/ramclustr_csv.xml
+++ b/tools/ramclustr/ramclustr_csv.xml
@@ -18,7 +18,11 @@
             -e 'source("${__tool_directory__}/ramclustr_wrapper.R")'
             -e 'x <- ramclustr_csv(
                 ms="$ms_csv.ms",
-                idmsms="$ms_csv.idmsms",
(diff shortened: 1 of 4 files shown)


martenson:
Are the parameter changes unrelated? (i.e. only `if(!file.exists(idmsms))` is the fix?)

hechth:
> Are the parameter changes unrelated? (i.e. only `if(!file.exists(idmsms))` is the fix?)

The proper setting it to NULL if it is not valid is the fix. The changes are that 'st' is not optional in CSV. Sorry I should have separated these things.

martenson (approved):

PR #82, opened 2021-02-04
PubChem RDF tools

xtracko:
This PR enables downloading [PubChem RDF](https://pubchemdocs.ncbi.nlm.nih.gov/rdf) and building our compound database which can be later used in annotation.


martenson:
@xtracko Is there a possibility for adding a basic test?

xtracko:
I am working on adding the test right now. I will commit it shortly after.

martenson:
@xtracko could you please commit the test so we can move this forward? Thank you.

tools/pubchem/pubchem_build.xml
```diff
@@ -0,0 +1,36 @@
...
+        <requirement type="package" version="1.2.1">pandas</requirement>
+        <requirement type="package" version="4.0.30">pyodbc</requirement>
+        <requirement type="package" version="0.5.4">python-snappy</requirement>
+        <requirement type="package" version="3.8">python</requirement>
+    </requirements>
+
+    <command detect_errors="aggressive"><![CDATA[
+        tar -xvf ${pubchem_rdf}
+
+        python '${__tool_directory__}/pubchem_build.py'
```
martenson, line 16: Galaxy treats this as one command, hence the error `tar: unrecognized option '--workers' Try 'tar --help' or 'tar --usage' for more information.` when trying to run this. We could chain it with some logical operators like `&&`, extract to an external script, etc.

martenson:
The tool now runs without error. However it is still unclear to me with what data to test it, 4 random RDF sets I've chosen produce 0 bytes parquet dataset.

hechth:
What is the current status of this? Can this be merged, does it need further work? If so, what are the issues?

So that maybe we can move this a bit further :)

martenson:
@hechth fwiw it cannot be merged in this state since there are unanswered questions from my review.

hechth:
@martenson I think we can close this for now - in case it would be necessary, we can re-open.

martenson:
@hechth agreed, closed

PR #89, opened 2021-02-19
rewrite export tool to work with new thermo converter

martenson:
also abandon efforts to make it generic, added complexity is not worth it imo

The Thermo converter v1.3.2 does not detect outputs using the collection file discovery engine hence we do not have the information about the original filenames anymore. I took the opportunity to simplify the code and just allow the caller to specify the full path of the mzML to be created, the metadata file names are inferred from the same name.

### DIFF
--- a/tools/export_to_path/export_to_path.py
+++ b/tools/export_to_path/export_to_path.py
@@ -25,20 +25,25 @@ if not os.path.exists(export_dir):
 #  Make sure program will never write outside of the mounted target directory.
 if not os.path.commonpath([os.path.realpath(options.remote_path), SALLY_MOUNT_PREFIX]) == SALLY_MOUNT_PREFIX:
     raise Exception(f"Invalid export path supplied: {options.remote_path}")
-remote_paths = [options.remote_path]
-if len(args) > 1:
-    remote_paths.append(f"{options.remote_path[:-5]}.json")
-if len(args) == 3:
-    remote_paths.append(f"{options.remote_path[:-5]}.txt")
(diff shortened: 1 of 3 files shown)


tools/export_to_path/export_to_path.py
```diff
@@ -1,53 +1,44 @@
...
+if not os.path.exists(export_dir):
+    os.makedirs(export_dir)
+#  Make sure program will never write outside of the mounted target directory.
+if not os.path.commonpath([os.path.realpath(options.remote_path), SALLY_MOUNT_PREFIX]) == SALLY_MOUNT_PREFIX:
+    raise Exception(f"Invalid export path supplied: {options.remote_path}")
+remote_paths = [options.remote_path]
+if len(args) > 1:
+    remote_paths.append(f"{options.remote_path[:-5]}.json")
+if len(args) == 3:
+    remote_paths.append(f"{options.remote_path[:-5]}.txt")
```
hechth, line 32: Could you explain what these checks exactly are in a comment (maybe in the code)? To me they seem a bit magic from just having a quick look - why add txt if `len(args) == 3` and json if `len(args) > 1` ?

martenson: I agree it is not very robust, the idea was to have positional arguments that would be used with json metadata option as default but would also work when eventually we have thermo that outputs both.

I also have a branch that narrows this down to "copy one dataset to a given path". Since we control all the paths outside of Galaxy it might be the easier way forward.

PR #94, opened 2021-02-24
add tool to merge metadata with features data

maximskorik:
a tool that creates a proper input for WaveICA and RamClustR


tools/metadata_merger/metadata_merger.py
```diff
@@ -0,0 +1,56 @@
...
+        dataframe = pd.read_csv(args.dataframe, sep="\t", index_col=0)
+        dataframe = dataframe.transpose()
+    elif args.extension == "csv":
+        dataframe = pd.read_csv(args.dataframe, sep=",", index_col=0)
+    else:
+        print("Wrong format. Input has to be \"*.tsv\" or \"*.csv\" file.", file=sys.stderr)
+        return(1)
+
+    n = len(dataframe)
+    print("Reading features table of {} samples...".format(n), file=sys.stdout)
```
martenson, line 29: You're on python3, check out [f-strings](https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python) for easier formatting.

martenson: `f"Reading features table of {len(dataframe)} samples..."`

tools/metadata_merger/metadata_merger.xml
```diff
@@ -0,0 +1,38 @@
+<tool id="metadata_merger" name="Metadata Merger" version="0.0.1" python_template_version="3.5">
+    <description>merges metadata with features table</description>
+    <requirements>
+        <requirement type="package" version="3.8.5">python</requirement>
+        <requirement type="package" version="1.2.1">pandas</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        python $__tool_directory__/metadata_merger.py --dataframe "$features_table" --extension "${features_table.ext}" --metadata "$metadata_table" --output "$merged_data"
```
martenson, line 8: Please make sure the needed variables are single quoted, so they don't get expanded in the shell. (from [IUC standards](https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#command-formatting))

> All Cheetah variables for text parameters, input and output files must be single-quoted, e.g. '${var_name}'.

tools/metadata_merger/metadata_merger.xml
```diff
@@ -0,0 +1,38 @@
...
+            <param name="metadata_table" value="metadata-test.tsv" ftype="tsv"/>
+            <output name="merged_data" file="aplcms_converted-merged.csv" ftype="csv"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+        A simple tool to merge samples metadata with features table.
+        
+        **Workflow position**
+
+        - Upstream tools: `apLCMS to RamClustR converter <https://umsa.cerit-sc.cz/root?tool_id=testtoolshed.g2.bx.psu.edu/repos/recetox/aplcms_to_ramclustr_converter/aplcms_to_ramclustr_converter/0.0.1>`_ or `XCMS fillChromPeaks <https://umsa.cerit-sc.cz/root?tool_id=toolshed.g2.bx.psu.edu/repos/lecorguille/xcms_fillpeaks/abims_xcms_fillPeaks/3.6.1+galaxy0/>`_
```
martenson, line 34: The links could probably go to the development repositories, as a more stable target.

tools/metadata_merger/metadata_merger.xml
```diff
@@ -0,0 +1,38 @@
...
+
+        - Downstream tools: `RamClustR <https://umsa.cerit-sc.cz/root?tool_id=testtoolshed.g2.bx.psu.edu/repos/recetox/ramclustr_csv/ramclustr_csv/1.1.0+galaxy0>`_ (CSV) or `WaveICA <future_link>`_
```
martenson, line 36: future_link 🤣

tools/metadata_merger/metadata_merger.xml
martenson: The toolshed link is actually good, but we do not have that for our tools yet. :)

tools/metadata_merger/test-data/aplcms_converted-test.csv
```diff
@@ -0,0 +1,7 @@
+sample_name,100.01124231464279_316.6324148139556,100.0140326116781_370.3911481191968,100.03941952410403_150.6579990293048,100.07581095178341_53.375130879368484
+VT_160120_002,29239096.399301223,40207085.08482562,3031215.6621509064,289190890.42614394
+VT_160120_004,9373024.38821382,39892591.38643598,1156692.2274786886,314872910.16817033
+VT_160120_006,11575595.878938954,41211883.00982785,1188030.695089618,308427518.55757684
+VT_160120_008,26277808.30956555,45866485.05309503,2255800.536242036,119045892.73239136
+dataset_4692,24658120.59201486,38960230.26046665,4663513.710815267,144921892.4161568
```
martenson, line 6: What do these lines starting with `dataset_` mean?

martenson (commented):
I think we should look into already existing tools like [join](https://umsa.cerit-sc.cz/root?tool_id=toolshed.g2.bx.psu.edu/repos/bgruening/text_processing/tp_easyjoin_tool/1.1.2) and make sure we cannot use existing software.

tools/metadata_merger/test-data/metadata-test.tsv
```diff
@@ -0,0 +1,6 @@
+sample_name	class   sampleType  injectionOrder  batch
```
martenson, line 1: this is not a valid TSV file
Screenshot_3_1_21__1_05_PM

tools/metadata_merger/metadata_merger.py
maximskorik: That looks much better, somehow never saw that type of formatting.

tools/metadata_merger/metadata_merger.xml
maximskorik: Well, I wanted to put there some placeholder just to be sure if I use valid xml-schema. Eventually, there will be a link though 😄

tools/metadata_merger/test-data/aplcms_converted-test.csv
maximskorik: These are a few samples that don't have corresponding metadata and should not be included in the output. Just to check if the join is performed correctly.

tools/metadata_merger/test-data/metadata-test.tsv
maximskorik: I noticed that there was something wrong with that but thought it's just a problem with interpreter on my machine. @hechth, that's the one that was in the Galaxy history you shared with me. Maybe there is some problem with the way we generate tables with metadata? Or it has just corrupted for some reason...

maximskorik:
> I think we should look into already existing tools like [join](https://umsa.cerit-sc.cz/root?tool_id=toolshed.g2.bx.psu.edu/repos/bgruening/text_processing/tp_easyjoin_tool/1.1.2) and make sure we cannot use existing software.



> I think we should look into already existing tools like [join](https://umsa.cerit-sc.cz/root?tool_id=toolshed.g2.bx.psu.edu/repos/bgruening/text_processing/tp_easyjoin_tool/1.1.2) and make sure we cannot use existing software.

You're totally right on this. There's a tool that does exactly the same. I've made an issue on this #97, if you could check it please.

martenson:
On the XCMS test case we can see rounding performed (see `954938.1313372459` rounded to `954938.131337246` in the result). We should make sure it is desired.

maximskorik:
> On the XCMS test case we can see rounding performed (see `954938.1313372459` rounded to `954938.131337246` in the result). We should make sure it is desired.

I mentioned this at the (25.2) meeting. It's unlikely we even have instruments that measure with such precision, hence I assumed it shouldn't be a problem.

maximskorik:
> On the XCMS test case we can see rounding performed (see `954938.1313372459` rounded to `954938.131337246` in the result). We should make sure it is desired.

I think it would be good if I write it down in the help section, what do you think?

martenson:
I think that unless we have some further requirements we should use the existing Transpose, and Join tools with an additional benefit of avoiding the unexpected rounding.

maximskorik:
> I think that unless we have some further requirements we should use the existing Transpose, and Join tools with an additional benefit of avoiding the unexpected rounding.

I just saw your comments on the issue #97 after I replied here. I agree that we should use the existing ones.

martenson:
Since this seems plausible I think we agree to use the existing tools. Please feel free to reopen if needed.

PR #95, opened 2021-02-25
Initial push for RMassBank tool

trachtok:
RMassBank workflow, includes creating of user-customizable settings file for RMassBank (template of this settings file is part of the tool) and both parts of RMassBank workflow. Apart from settings file template, two Python scripts ar also included. One adjusts the whole settings file based on parameters given by user in Galaxy (simply looks for a line that contains specific parameters and then re-writes the line with the user's parameter value), the second script just changes OpenBabel path in case the user wants to run RMassBank with custom settings file (because not all parameters in settings file can be wrapped in XML for Galaxy, some more complex can be edited only locally by user). Unfortunately, RMassBank right now has a hard-coded executable 'babel' (newest OpenBabel 3.0.0 was renamed to 'obabel') and requires a full path to it, which is a problem when using OpenBabel in conda env in Galaxy. Therefore, the actual path to OpenBabel has to be resolved on the run-time of Galaxy tool and passed as an argument for the second Python script which it then writes into the settings file.

Also, because RMassBank requires input files to have specific suffixes (not .dat of Galaxy), all input files have to be symlinked to original names. 

I plan to write a GitHub issue to authors of RMassBank summarizing all these small problems and hopefully, they will adjust their code so that at least the second Python script will not be necessary.


martenson:
The test is failing because you are using `zip` in your command but did not declare it as a requirement. 

The tool test result, ahich usually has the reason for failing, is always available after the run for download: https://github.com/RECETOX/galaxytools/actions/runs/600110002

martenson:
Looks good. Please make sure the needed variables are single quoted, so they don't get expanded in the shell. (from [IUC standards](https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#command-formatting))

> All Cheetah variables for text parameters, input and output files must be single-quoted, e.g. '${var_name}'.

trachtok:
Thanks @martenson, I will add the single quotes, although the code seems to work just fine without them. I went through the IUC standards and some tools published on Galaxyproject and found out that they write it either as `$variable` or `'$variable'` or `'${variable}'` so I am slightly confused about what is right (and planemo test takes all options and finishes tests successfully).

martenson:
@trachtok It is a best practice, not a requirement, or something that would always fail. We never want to expand those variables in shell, which would under some circumstances happen, and this will prevent it.

martenson:
Thank you @trachtok for making the changes. 17 MB test file is still pretty big, but hopefully it provides a good test case and we can reuse it in other tools.

Issue #35, opened 2020-11-05
RAMClustR: enhancements


hechth:
## Description
Add a galaxy tool wrapping RAMClustR.
The tool can be run with XCMS objects or CSV tables as input. These are handled in two different tools for ease of use, but share the same code/container.

### General
- [x] Forward parameters to script from galaxy environment for `xcmsSet` object use
- [x] Implement container to run RAMClustR
- [x] Split tool into two separate parts, one for XCMS object and one for CSV input
- [x] ~~Switch from container to bioconda package once available~~ (continues in https://github.com/RECETOX/galaxytools/issues/114)

### Interface Galaxy -> Tool
- [x] Write wrapper script in tool directory which transforms and handles input - ideally an R script so that the Galaxy tool only calls the function in the R script and we have an interface between the RAMClustR libray and the Galaxy tool

Handle `batch`, `qc` and `order` input parameters by taking a metadata CSV similar to XCMS & WaveICA
- [x] Find proper data type for input - some type of array, text file, numpy array, R array etc
- [x] Transform to R vectors and pass to tool

### Define Experiment Tool
- [x] #216

### CSV Tool
- [x] Implement parameter handling/forwarding using script in tool directory

### Tests:
- [x] Basic test for XCMS based tool
- [x] Basic test for csv based tool once #72  is resolved

### Documentation
- [x] Add upstream/downstream tools
- [x] Add basic background information
- [ ] Describe outputs

martenson:
I made PRs for conda packaging of ramclustr (https://github.com/bioconda/bioconda-recipes/pull/25270) and its last non-packaged dependency (https://github.com/bioconda/bioconda-recipes/pull/25267). When they get merged and built we should switch our wrappers to use them instead of container.

hechth:
The container installs directly from github, so it's also not a bad solution IMHO.

martenson:
If the overhead of maintaining a container can be avoided I'd go for conda. This also gives us very lightweight biocontainers/singularity images for free.

hechth:
> If the overhead of maintaining a container can be avoided I'd go for conda. This also gives us very lightweight biocontainers/singularity images for free.

This is definitely true and also takes away the burden of versioning ;)

hechth:
@martenson I tried switching to using the bioconda package by changing the requirement from
```xml
    <requirements>
        <container type="docker">recetox/ramclustr:1.1.0-recetox0</container>
    </requirements>
```
to
```xml
    <requirements>
[...]

PR #121, opened 2021-05-19
RAMClustR: Added handling of metadata

hechth:
Contributes to #35 as it includes handling of the `qc`, `batch` and `order` parameters. The added input is a metadata csv file similar to the one used in XCMS which contains this data and which can be added as an optional input.


maximskorik (approved):

martenson:
updated on UMSA

PR #133, opened 2021-06-16
Switch to our aplcms fork, use parquet for output

xtracko:
Adapts the Galaxy tools to changes introduced by recetox/recetox-aplcms#14


martenson:
We're in a pickle re:`@TOOL_VERSION@`, if we update it to our package version (v0.9.2) then Galaxy will treat it as older than our current version (v6.6.6). Maybe we should switch the tool id and start a new toolshed repository for it.

hechth:
@martenson @xtracko So are we making the recetox-aplcms a new Galaxy tool? With the new versioning, maintenance and authorship, I think that would be just consistent.

martenson:
Sounds good to me, I pushed the commits that make this happen to this PR.

hechth (commented):
Seems good.

martenson:
I renamed the folder to use underscore as suggested by @xtracko

xtracko:
@martenson The whole thing fails because of some problems with R packages (visibility of some functions outside the package). So first, we must re-release recetox-aplcms conda package...

martenson:
The approach to using the main.R wrapper does not seem to be correct yet.

```
Error in library(recetox.aplcms) : 
  there is no package called ‘recetox.aplcms’
Calls: source -> withVisible -> eval -> eval -> library
Execution halted
```

martenson:
With the proper package name now we are finally getting somewhere

```
**** feature extraction ****
**** time correction ****
**** performing time correction ****
m/z tolerance level:  1.13783763305855e-05
time tolerance level: 528.630506001489
the template is sample 1
**** feature alignemnt ****
**** performing feature alignment ****
m/z tolerance level:  1.13783763305855e-05
time tolerance level: 528.666449313702
**** weaker signal recovery ****
Error: IOError: Failed to open local file 'extracted-/tmp/tmpu50kf5iy/files/9/9/c/dataset_99c28e91-c998-4c35-88c7-530f62f1b693.parquet'. Detail: [errno 2] No such file or directory
Execution halted
```

tools/recetox_aplcms/main.R
```diff
@@ -0,0 +1,72 @@
...
+  arrow::read_parquet(filename, col_select = known_table_columns())
+}
+
+save_pairing <- function(df, filename) {
+  write.table(df, filename, row.names = FALSE, col.names = c("new", "old"))
+}
+
+save_extracted_features_as_collection <- function(dfs, filenames) {
+  filenames <- tools::file_path_sans_ext(filenames)
+  filenames <- paste0("extracted-", filenames, ".parquet")
```
martenson, line 36: Error: IOError: Failed to open local file 'extracted-/tmp/tmpu50kf5iy/files/9/9/c/dataset_99c28e91-c998-4c35-88c7-530f62f1b693.parquet'. Detail: [errno 2] No such file or directory

martenson: this has to work on something else than the actual given filepath, for the sake of collection making we could start a a filename pattern and then use `<discover_datasets>`

tools/recetox_aplcms/recetox_aplcms_hybrid.xml
```diff
@@ -67,9 +67,9 @@
 
     <tests>
         <test>
-            <param name="files" value="mbr_test0.mzml,mbr_test1.mzml,mbr_test2.mzml" ftype="mzml"/>
-            <param name="known_table" value="known_table.parquet" ftype="parquet"/>
-            <output name="recovered_feature_sample_table" file="recovered_feature_sample_table.parquet" ftype="parquet" compare="sim_size" delta="1000"/>
+            <param name="known_table" value="known_table.parquet" ftype="parquet" />
+            <param name="files" value="mbr_test0.mzml,mbr_test1.mzml,mbr_test2.mzml" ftype="mzml" />
+            <output name="recovered_feature_sample_table" file="hybrid_recovered_feature_sample_table.parquet" ftype="parquet" compare="sim_size" delta="1000" />
```
hechth, line 78: Is there any other way than comparing with similar size, since it actually doesn't check the content of the file?

hechth (commented):
Looks good, maybe there is a way to make the tests more precise?

tools/recetox_aplcms/recetox_aplcms_hybrid.xml
martenson: I'd like to keep this PR to a minimum, so I can run performance tests over the weekend. So yes, but let's implement it in a subsequent PR.

hechth: Ok, it would be simply about the content. I'll push it in a new issue.

martenson:
The failing test should pass once ran against the updated master of this repo (needed newer Galaxy version to tests against)

martenson (approved):

PR #135, opened 2021-06-19
aplcms: change the requirement to r-base package

martenson:


martenson:
namely this should fix this: https://github.com/galaxyproject/planemo-monitor/runs/2866757395?check_suite_focus=true#step:6:508

PR #147, opened 2021-07-26
Ramclustr input unification

xtrojak:
Combined tools **RAMClustR (XCMS)** and **RAMClustR (CSV)** into a single tool **RAMClustR** with conditional parameter allowing the user to select input format.


hechth (commented):
Looks good, didn't run manual planemo tests though, but CI should do that.

martenson (approved):

martenson:
Just to clarify, is this the dockerfile we use as requirement @hechth ? https://github.com/RECETOX/RAMClustR/blob/hechth/issue1/Dockerfile

hechth:
Yes!

PR #187, opened 2021-11-03
WaveICA: add metadata to output and fix column names

### LINKED ISSUE #179: waveica: Improve documentation
- [ ] The tool documentation should indicate what are the expected columns and show some rows of an [example](https://github.com/RECETOX/galaxytools/blob/master/tools/waveica/test-data/features-test.csv) table from the test data.
- [ ] The tool should mention that the data mustn't contain missing values which have to be filled with 0
- [ ] The individual parameters should be described in more detail.

### LINKED ISSUE #186: waveica: output table doesn't contain metadata and has altered column names
During normalization, the tool removes the metadata columns. They should be added back. Also the column names shouldn't be changed.

maximskorik:
* removed `data_wave` prefix from the output column names
* added check for the presence of metadata in input
* added metadata columns to the output table
* added verbal error when input contains `NA` values
* changed samples ordering to account for repeating numbering
* added example input to help section and improved documentation

Update 5.11.:
* added a footnote with reference to R `wavelet::wt.filter` function, used in `WaveICA`.
* updated `alpha` parameter description, clarifying that the parameter influences a trade-off between spatial and temporal ICA. It's not clear which value ([0,1]) corresponds to which ICA. You can check how the `alpha` influences independent component analysis [here](https://github.com/dengkuistat/WaveICA/blob/master/R/unbiased_stICA.R) if you're into linear algebra.

Closes #186, closes #179


tools/waveica/waveica_wrapper.R
```diff
@@ -38,7 +38,7 @@ Check that the following columns are present in your dataset: [sampleName, class
         alpha = alpha
         )
 
-    data[, -c(1:5)] <- features
+    data[, -c(1:5)] <- features$data_wave
```
hechth, line 45: I think it would be more robust to include and exclude the columns by name instead of by index.

tools/waveica/waveica_wrapper.R
```diff
@@ -10,25 +10,28 @@ waveica <- function(
...
+        stop("Error: dataframe cannot contain NULL values!
+Make sure that your dataframe does not contain empty cells")
+    }
+    else if (!all(required_columns %in% colnames(data))) {
+        stop("Error: missing metadata!
+Make sure that the following columns are present in your dataframe: [sampleName, class, sampleType, injectionOrder, batch]")
+    }
 
     # sort data by injection order
     data <- data[order(data$injectionOrder, decreasing = FALSE), ]
```
hechth, line 26: Does this injection order also reflect the batch number? Because I assume that the ordering should be batch > injectionOrder, otherwise all first samples of all batches will be on the top and all last samples on the bottom etc.

Or is this the desired behaviour?

tools/waveica/waveica_wrapper.R
```diff
@@ -10,25 +10,28 @@ waveica <- function(
...
-
-    # remove blanks from dataset
-    if (exclude_blanks) {
-        data <- exclude_group(data)
-    }
-
     # separate data into features, batch and group
-    features <- data[, -c(1:4)]
-    group <- as.numeric(data$class)
+    features <- data[, -c(1:5)]
```
hechth, line 29: Again, selecting by name might be more robust.

tools/waveica/waveica_wrapper.R
maximskorik: I agree, fixed in 2413166

tools/waveica/waveica_wrapper.R
maximskorik: Fixed in 2413166.

tools/waveica/waveica_wrapper.R
maximskorik: This is an excellent point! I assumed that the injection order would be batch-independent, i.e., the numbering would go `from 1 to the number of samples` as it was in the test data of the `WaveICA` package. But it wouldn't work if the numbering repeated for every batch. I accounted for that in 49f4b2a.

hechth:
The galaxy tool still outputs row names and quotes the string type entries, which shouldn't be the case.

maximskorik:
> The galaxy tool still outputs row names and quotes the string type entries, which shouldn't be the case.

Fixed in ee079a2

tools/waveica/waveica_wrapper.R
```diff
@@ -83,7 +95,7 @@ exclude_group <- function(data) {
 
 
 # Store output of WaveICA in a tsv file
-store_data <- function(normalized_data, output) {
-    write.table(normalized_data, file = output, sep = "\t", col.names = NA)
+store_data <- function(data, output) {
+    write.table(data, file = output, sep = "\t", col.names = NA, quote = FALSE)
```
hechth, line 99:
```suggestion
    write.table(data, file = output, sep = "\t", row.names = FALSE, quote = FALSE)
```

tools/waveica/test-data/incomplete_metadata_data.csv
```diff
@@ -0,0 +1,6 @@
+sampleName,class,sampleType,batch,M85T34,M86T41,M86T518,M86T539
+1,VT_160120_002,sample,sample,1,228520.06430737,35646729.2154397,2386896.97966461,1026645.83653468
+2,VT_160120_004,sample,sample,1,90217.384387202,35735702.457216,2456290.69621518,1089246.46040563
+3,VT_160120_006,sample,sample,1,235656.752883839,37021134.4527116,8873450.40260241,837856.449608585
+4,VT_160120_008,sample,sample,1,16622.9351783435,44302499.262606,2466946.89667101,994979.069689685
+5,VT_160120_010,sample,sample,1,62385.0742465736,44639738.0735709,2389372.85729467,954938.131337246
```
hechth, line 6:
```suggestion
sampleName,class,sampleType,batch,M85T34,M86T41,M86T518,M86T539
VT_160120_002,sample,sample,1,228520.06430737,35646729.2154397,2386896.97966461,1026645.83653468
VT_160120_004,sample,sample,1,90217.384387202,35735702.457216,2456290.69621518,1089246.46040563
VT_160120_006,sample,sample,1,235656.752883839,37021134.4527116,8873450.40260241,837856.449608585
VT_160120_008,sample,sample,1,16622.9351783435,44302499.262606,2466946.89667101,994979.069689685
VT_160120_010,sample,sample,1,62385.0742465736,44639738.0735709,2389372.85729467,954938.131337246
```

tools/waveica/test-data/normalized_data.tsv
```diff
@@ -0,0 +1,6 @@
+	sampleName	class	sampleType	injectionOrder	batch	M85T34	M86T41	M86T518	M86T539
+1	VT_160120_002	sample	sample	1	1	355200.506508035	75115889.9077485	6101488.54615418	2007379.02604984
+2	VT_160120_004	sample	sample	2	1	216897.826587868	75204863.1495248	6170882.26270475	2069979.64992079
+3	VT_160120_006	sample	sample	3	1	362337.195084504	76490295.1450204	12588041.969092	1818589.63912375
+4	VT_160120_008	sample	sample	4	1	143303.377379009	83771659.9549148	6181538.46316058	1975712.25920485
+5	VT_160120_010	sample	sample	5	1	189065.516447239	84108898.7658797	6103964.42378424	1935671.32085241
```
hechth, line 6:
```suggestion
sampleName	class	sampleType	injectionOrder	batch	M85T34	M86T41	M86T518	M86T539
VT_160120_002	sample	sample	1	1	355200.506508035	75115889.9077485	6101488.54615418	2007379.02604984
VT_160120_004	sample	sample	2	1	216897.826587868	75204863.1495248	6170882.26270475	2069979.64992079
VT_160120_006	sample	sample	3	1	362337.195084504	76490295.1450204	12588041.969092	1818589.63912375
VT_160120_008	sample	sample	4	1	143303.377379009	83771659.9549148	6181538.46316058	1975712.25920485
VT_160120_010	sample	sample	5	1	189065.516447239	84108898.7658797	6103964.42378424	1935671.32085241
```

hechth (approved):

PR #193, opened 2021-11-12
waveica: change dependencies to Conda

maximskorik:
* replace Docker dependency with Bioconda **recetox-waveica** package


martenson:
nice @maximskorik !

PR #196, opened 2021-12-02
waveica: add WaveICA 2.0

maximskorik:
+ added **WaveICA (single batch)** galaxy tool that implements WaveICA 2.0 functionality
+ updated both tools to depend on **r-recetox-waveica v0.2.0** conda package
+ adjusted R wrapper to be compatible with both tools
+ updated `.shed.yml`
+ added test-data and documentation to **WaveICA (single batch)**
+ minor updates to **WaveICA** documentation
+ added macros to share identical parameters/features between the tools

**Upd.**:
**WaveICA (single batch)** tool produced dataframes with insignificantly different intensities depending on whether `planemo test` was executed with `--muled_containers` parameter or without. I don't know what exactly caused that, but I had to change the test data slightly to pass the tests (see 7677bb7). This shouldn't be a problem since the test data with only five samples are not entirely representative anyway and is there only to make sure that I/O and the wrapper function properly.


hechth:
Naming should be eventually adopted according to RECETOX/WaveICA#9

martenson (commented):
Have you considered consolidating these two overlapping tools into one that has two modes? Something like default/singlebatch -- similar to what can be seen e.g. at https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/iuc/merqury/merqury/1.3+galaxy1

maximskorik:
> Have you considered consolidating these two overlapping tools into one that has two modes? Something like default/singlebatch -- similar to what can be seen e.g. at https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/iuc/merqury/merqury/1.3+galaxy1

No, I have not. This looks like a neat alternative. @hechth, what would you say?

hechth:
I mean we have something very similar already with the RAMClustR program that can either consume directly an `xcmsExperiment` object or a `csv`, so this would be something similar.

hechth (approved):

hechth:
@martenson according to me this can be merged - what do you think?

martenson:
@hechth @maximskorik I think this looks very good. Nice job Maxim.

PR #203, opened 2022-01-06
188 tabular formats

### LINKED ISSUE #188: matchms formatter: Accept all tabular types as input
This also means removing the hardcoded \t delimiter and using a more sophisticated approach (https://stackoverflow.com/questions/45732459/retrieve-delimiter-infered-by-read-csv-in-pandas)

xtrojak:
PR to support multiple tabular formats in `matchms_formatter`.
For now, `.tsv` and `.csv` formats are allowed.
This required to change both Galaxy tool wrapper to accept these formats and reading of the fils using `pandas.read_csv` function.

Close #188.

### DIFF
--- a/tools/matchms/matchms_formatter.xml
+++ b/tools/matchms/matchms_formatter.xml
@@ -31,8 +31,8 @@
     </configfiles>
 
     <inputs>
-        <param label="Scores Table" name="scores" type="data" format="tsv" help="Scores output table." />
-        <param label="Matches Table" name="matches" type="data" format="tsv" help="Scores output table." />
+        <param label="Scores Table" name="scores" type="data" format="table,csv,tsv" help="Scores output table." />
+        <param label="Matches Table" name="matches" type="data" format="table,csv,tsv" help="Scores output table." />
 
         <conditional name="method">
             <param name="selection" type="select" label="Formatting method">
@@ -52,7 +52,7 @@
 
     </inputs>
     <outputs>
-        <data label="${tool.name} (${method.selection}) on ${on_string}" name="output" format="tsv" />
+        <data label="${tool.name} (${method.selection}) on ${on_string}" name="output" format_source="scores" metadata_source="scores" />
     </outputs>
 
     <tests>
--- a/tools/matchms/matchms_formatter.xml
+++ b/tools/matchms/matchms_formatter.xml
@@ -31,8 +31,8 @@
     </configfiles>
 
     <inputs>
-        <param label="Scores Table" name="scores" type="data" format="table,csv,tsv" help="Scores output table." />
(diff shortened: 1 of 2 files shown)


hechth:
Regarding the formats this should be all for now.

martenson:
`tabular` is a valid format (included in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/datatypes_conf.xml.sample#L413 and implemented at https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/tabular.py#L40) 

The format itself is widely used and I think Galaxy will also sniff the delimiter during job/upload processing.

tools/matchms/matchms_formatter.xml
```diff
@@ -52,7 +52,7 @@
 
     </inputs>
     <outputs>
-        <data label="${tool.name} (${method.selection}) on ${on_string}" name="output" format="tsv" />
+        <data label="${tool.name} (${method.selection}) on ${on_string}" name="output" format_source="scores" metadata_source="scores" />
```
martenson, line 55: What sort of metadata are we looking to source from the input file here?

tools/matchms/matchms_formatter.xml
```diff
@@ -31,8 +31,8 @@
     </configfiles>
 
     <inputs>
-        <param label="Scores Table" name="scores" type="data" format="tsv" help="Scores output table." />
-        <param label="Matches Table" name="matches" type="data" format="tsv" help="Scores output table." />
+        <param label="Scores Table" name="scores" type="data" format="table,csv,tsv" help="Scores output table." />
```
martenson, line 34: I think this is fine but I am not clear on how `table` behaves -- as you mentioned it is not a tabular format in Galaxy so it'd make sense to drop it.

tools/matchms/matchms_formatter.xml
xtrojak: You are right, it doesn't make sense in tabular type of data (fixed in 45ab1df).

tools/matchms/matchms_formatter.xml
xtrojak: I have removed `table` format in 45ab1df.

hechth:
@xtrojak how does the tool behave if you supply the actual `tabular` type?

xtrojak:
> @xtrojak how does the tool behave if you supply the actual `tabular` type?

It seems to be working fine just like `tsv`.

hechth:
@martenson wouldn't it make more sense to just let the tool operate on the `tabular` type? Or is this not possible

hechth (approved):

martenson:
@hechth I think it still does since it is a superclass, just with an implicit conversion:
Galaxy

martenson:
@hechth PR here: https://github.com/RECETOX/galaxytools/pull/206

PR #204, opened 2022-01-07
194 mgf support

### LINKED ISSUE #194: matchms: Add mgf support for reading spectra
Currently we only support reading MSP, we could also add the mgf reader and writer which would improve the versatility of the tool.

xtrojak:
This PR adds `.mgf` format support for input spectra files in `matchms` wrapper.

Close #194.


hechth:
@xtrojak Can you pull my commit into your branch to append it to the PR?

tools/matchms/matchms_wrapper.py
```diff
@@ -17,21 +17,28 @@ def main(argv):
...
-    parser.add_argument("output_filename_matches", type=str, help="Path where to store the output .csv matches.")
+    parser.add_argument("output_filename_scores", type=str, help="Path where to store the output .tsv scores.")
+    parser.add_argument("output_filename_matches", type=str, help="Path where to store the output .tsv matches.")
     args = parser.parse_args()
 
-    queries_spectra = list(load_from_msp(args.queries_filename))
+    try:
+        queries_spectra = list(load_from_msp(args.queries_filename))
+    except ValueError:
+        queries_spectra = list(load_from_mgf(args.queries_filename))
```
hechth, line 33: this doesn't really look like a good solution! Maybe we can add the file type as a flag to the wrapper and then call the respective function based on the file type? This could also be done via a dictionary, mapping the extension to the loader function.

hechth (changes requested):

tools/matchms/matchms_wrapper.py
xtrojak: I agree, although I don't know how do it without asking user to explicitly choose the file type and then the file itself, which seems laborious to me.

I think the file extension cannot be used, because Galaxy is using `.dat` files for all data internally.

hechth: > It can be used, using the ${var.ext} tag, see https://github.com/RECETOX/galaxytools/blob/e4cdacabd9d58524d3271e005644b8f2d56d83fc/tools/riassigner/riassigner.xml#L12

xtrojak: That is a cool feature, I didn't know about it. Thanks! Fixed in 0c7ba48.

hechth (approved):

PR #206, opened 2022-01-14
matchms: add tabular datatype explicitly

martenson:


xtrojak:
If `tsv` is a subclass of `tabular`, I suppose one of them is redundant?

martenson:
It avoids the implicit conversion (tsv->csv) which I suppose is what @hechth requested in https://github.com/RECETOX/galaxytools/pull/203#issuecomment-1012911410

PR #207, opened 2022-01-14
189 separate matchms filtering functionality

### LINKED ISSUE #165: matchms: add m/z range filter option to tool
Add the option to use the [select by mz](https://github.com/matchms/matchms/blob/master/matchms/filtering/select_by_mz.py) filter to limit the mass spectral library to exclude peaks outside the selected mass range.

### LINKED ISSUE #189: matchms: Extract filtering into specific tool
The matchms filter modules should be extracted into an individual tool with the following filters:

- [ ] Normalize intensities
- [ ] Apply default filters
- [ ] Filter relative intensity
- [ ] Filter m/z range
- [ ] Clean metadata -> appluy all "add" and cleaning filters if possible, so that the spectra have canonical metadata

The galaxy tool could work in such a way that it allows iteratively adding of steps in the filtering and the tool just works through these steps, or to always only perform a single step, which will on the other hand result in very long pipelines.

### LINKED ISSUE #205: matchms: update version
Update to latest matchms version.

xtrojak:
Implemented several `matchms` filtering methods (enumerated in #189) as a separate wrapper. These were implemented as a series of conditionals (with possible arguments), assuming every filter can be applied only once and their order is fixed (i.e. the order of filters is not significant). Removed all filtering from `similarity` wrapper (the original one, just renamed). Also updated version of `matchms` to `0.11.0`.

Close #189.
Close #205.
Close #165.


tools/matchms/matchms_similarity_wrapper.py
```diff
@@ -12,10 +11,22 @@
 from pandas import DataFrame
 
 
+def convert_precursor_mz(spectrum):
+    """
+    Precursor m/z must be present for ModifiedCosine similarity metric
+    and it needs to be converted to float.
+    """
```
martenson, line 18:
```suggestion
    Check the presence of precursor m/z since it is needed for ModifiedCosine similarity metric. Convert to float if 
    needed, raise error if missing.
    """

```

martenson: I believe active voice makes for a more readable code documentation. Here is a rephrasing suggestion that carries the same information but explicitly describes the steps taken inside the method.

tools/matchms/matchms_filtering.xml
```diff
@@ -0,0 +1,153 @@
+<tool id="matchms_filtering" name="matchMS filtering" version="0.11.0+galaxy0">
+    <description>calculate the similarity score and matched peaks</description>
+    <requirements>
+        <requirement type="package" version="0.11.0">matchms</requirement>
```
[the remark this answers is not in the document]
xtrojak: Because as a part of this PR, we are also updating the version #205.

xtrojak:
I don't know how to ensure that at least one `conditional` is checked. It is possible to run the tool without any filters, which basically just duplicates the file. Alternatively, this could lead to an exception on the wrapper side.

tools/matchms/matchms_filtering.xml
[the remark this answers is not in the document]
martenson: oh, I missed the renaming of `tools/matchms/matchms.xml → tools/matchms/matchms_similarity.xml `

tools/matchms/matchms_filtering.xml
```diff
@@ -0,0 +1,153 @@
...
+
+        <conditional name="default_filters">
+            <param name="is_true" label="Apply default filters" type="boolean" truevalue="TRUE" falsevalue="FALSE" checked="false"
+                   help="Collection of filters that are considered default and that do no require any (factory) arguments."/>
+        </conditional>
+
+        <conditional name="clean_metadata">
+            <param name="is_true" label="Clean metadata" type="boolean" truevalue="TRUE" falsevalue="FALSE" checked="false"
+                   help="Apply all adding and cleaning filters if possible, so that the spectra have canonical metadata." />
+        </conditional>
```
martenson, line 60: I am not sure the first three checkboxes need to have a conditional. There is nothing on the toolform being triggered by the selection, could these be just checkboxes?

xtrojak: Yes, you are right, the reason is purely aesthetic so that all the filters looks the same. Also, to the best of my knowledge, you can't add a `help` section to a checkbox option.

xtrojak:
> @xtrojak What about applying all filters when none is selected? If that is not preferred solution we can throw a quick exception/error from the command. Nevertheless the tool should not execute if we can determine that it will simply create a dupe.

I think applying all filters is not an option because there are no meaningful default values for `intensity` and `m/z` range filters. In 5a4cf59 I added an exception raise when no filter is provided.

martenson (approved):
>  I think the minimalisation of tests is topic for another issue, but I can do it as a part if this PR.

My point is that every megabyte added to a git repo is stuck in its history forever. But if there is no way to trim the file and still make the test useful as before I am fine with this one getting in.

PR #227, opened 2022-03-03
Implemented basic RI filtering via matchms

### LINKED ISSUE #226: matchms: Implement retention index filtering
With matchms 0.14 being available via bioconda, we can implement the RI filtering using the new [MetadataMatch](https://github.com/matchms/matchms/blob/master/matchms/similarity/MetadataMatch.py) function.

Easiest would be to simply have a toggle whether to also match RI or not and if it is set, set the RI tolerance.

If the RI match is below tolerance, the score and number of matches could be set to 0, or we could output a third matrix with the RI scores which is also pushed to the formatter - or we add it as RI score to the `scores` object and change the formatter to also include the RI difference/match in the output table.

hechth:
Closes #226 

Implements basic RI filter which currently sets all scores to 0 if the RI is not in specified tolerance.
Updated matchms to 0.14.
Added test for new functionality and updated testsdata for new filter etc. from version 0.14


xtrojak (approved):
Looks good! Are we planning to implement more metadata filters? Then this implementation can be probably generalised.

martenson (approved):

Issue #225, opened 2022-02-24
biotransformer test failing on UMSA


martenson:
this seems to be related to issue described here https://github.com/RECETOX/galaxytools/pull/12#issuecomment-698943700

reported by @ElliottJP 

traceback when running on UMSA with test data:

```
Error loading JNI InChI native code.
You may need to compile the native code for your platform.
See http://jni-inchi.sourceforge.net for instructions.

Exception in thread "main" org.openscience.cdk.exception.CDKException: Unable to load native code; net.sf.jnati.NativeCodeException: Error creating directory: ?/.jnati/repo/jniinchi/1.03_1/LINUX-AMD64
	at org.openscience.cdk.inchi.InChIGeneratorFactory.<init>(InChIGeneratorFactory.java:87)
	at org.openscience.cdk.inchi.InChIGeneratorFactory.getInstance(InChIGeneratorFactory.java:102)
	at biotransformer.btransformers.Biotransformer.<init>(Biotransformer.java:130)
	at biotransformer.btransformers.ECBasedBTransformer.<init>(ECBasedBTransformer.java:56)
	at biotransformer.utils.HumanSuperBioTransformer.<init>(HumanSuperBioTransformer.java:47)
	at executable.BiotransformerExecutable.main(BiotransformerExecutable.java:997)
Caused by: net.sf.jniinchi.LoadNativeLibraryException: net.sf.jnati.NativeCodeException: Error creating directory: ?/.jnati/repo/jniinchi/1.03_1/LINUX-AMD64
	at net.sf.jniinchi.JniInchiWrapper.loadLibrary(JniInchiWrapper.java:98)
	at org.openscience.cdk.inchi.InChIGeneratorFactory.<init>(InChIGeneratorFactory.java:85)
	... 5 more
Caused by: net.sf.jnati.NativeCodeException: Error creating directory: ?/.jnati/repo/jniinchi/1.03_1/LINUX-AMD64
	at net.sf.jnati.deploy.repository.LocalRepository.createArtefact(LocalRepository.java:91)
	at net.sf.jnati.deploy.NativeArtefactLocator.deployArtefactToLocalRepository(NativeArtefactLocator.java:127)
	at net.sf.jnati.deploy.NativeArtefactLocator.getArtefact(NativeArtefactLocator.java:71)
	at net.sf.jnati.deploy.NativeLibraryLoader.loadLibrary(NativeLibraryLoader.java:47)
[...]

PR #235, opened 2022-03-18
update biotransformer to 3.0

martenson:
this works for me using conda dependencies, it would probably be worthwhile to test in a container


xtrojak (approved):
I verified that running tests in container works fine using

```planemo test --docker_sudo```

Thanks @hechth for recommending `--docker_sudo` parameter, as `--docker` does not work due to known [issue](https://github.com/RECETOX/galaxytools/issues/225) with permissions.

hechth:
@xtrojak so the tests all run fine?

hechth (approved):

xtrojak:
> @xtrojak so the tests all run fine?

Tests for biotransformer run fine, I didn't try other tests.

PR #236, opened 2022-03-18
upload all repos with green light to MTS

martenson:
This is done via minor galaxy version bump. That means it will also update the repos in TTS so we will have a point in time where these repos were in sync (sort of). This could be useful in the future.


martenson:
@xtrojak there is a MSME [test failing](https://github.com/RECETOX/galaxytools/actions/runs/2004919221), should we add some buffer like testing filesize or is this a legit fail?

xtrojak:
> @xtrojak there is a MSME [test failing](https://github.com/RECETOX/galaxytools/actions/runs/2004919221), should we add some buffer like testing filesize or is this a legit fail?

@martenson I suppose the reason is that IDSM service is currently down, which would lead to different results. We already started discussion about this test case [here](https://github.com/RECETOX/galaxytools/pull/199#issuecomment-1007481019), I don't know how to ensure the consistency of the results.

hechth:
@martenson Are tools which are not pushed to the MTS still pushed to the test toolshed so that we can test them on our Galaxy or is everything just pushed to MTS?

hechth:
@xtrojak @martenson After the update of MSME to 0.2.0 I think we should re-include it and see how the tests are running.

martenson:
@hechth tools in [skiplist](https://github.com/RECETOX/galaxytools/blob/master/.tt_skip) are completely ignored

martenson:
@hechth that has been already done in https://github.com/RECETOX/galaxytools/pull/241 -- the tests pass and it is being uploaded to MTS

PR #259, opened 2022-04-22
MSME matchms logging level

### LINKED ISSUE #255: MSME: Change log level to warning when reading files
Currently, matchms produces large amounts of print output in form of warnings during metadata hamronization. A way how to reduce this clutter would be to set the logging level to `ERROR` before reading the file to avoid the massive printouts.

xtrojak:
Set `matchms` logging level to `ERROR` to avoid extensive messages in stdout. This required to add `matchms` as MSME wrapper dependency.

Close #255.

### DIFF
--- a/tools/msmetaenhancer/msmetaenhancer.xml
+++ b/tools/msmetaenhancer/msmetaenhancer.xml
@@ -8,7 +8,6 @@
 
     <requirements>
         <requirement type="package" version="@TOOL_VERSION@">msmetaenhancer</requirement>
-        <requirement type="package" version="0.14.0">matchms</requirement>
     </requirements>
 
     <command detect_errors="exit_code"><![CDATA[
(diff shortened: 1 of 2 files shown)


tools/msmetaenhancer/msmetaenhancer.xml
```diff
@@ -8,6 +8,7 @@
 
     <requirements>
         <requirement type="package" version="@TOOL_VERSION@">msmetaenhancer</requirement>
+        <requirement type="package" version="0.14.0">matchms</requirement>
```
hechth, line 11: This is not necessary

tools/msmetaenhancer/msmetaenhancer_wrapper.py
```diff
@@ -16,6 +16,8 @@ def main(argv):
 
     app = Application(log_file=args.log_file)
 
+    # set matchms logging level to avoid extensive messages in stdout
+    set_matchms_logger_level("ERROR")
     # import .msp file
     app.load_spectra(args.input_file, file_format='msp')
```
hechth, line 23: The level can be returned back to `WARNING` after loading the spectra.

xtrojak: Good point, I set it back to `WARNING` before curation in [d3e8037](https://github.com/RECETOX/galaxytools/pull/259/commits/d3e8037f9e30a2fe2d25c6dafb890810a4aca0bb).

tools/msmetaenhancer/msmetaenhancer.xml
xtrojak: You are right, its installed with `MSME`.

hechth (approved):

PR #263, opened 2022-05-05
Tool to merge MSP files

wverastegui:
The tool reads spectrum of individual MSP files to merge them in a single MSP file.


tools/msp_merge/.shed.yml
```diff
@@ -0,0 +1,13 @@
+owner: recetox
+remote_repository_url: "https://github.com/RECETOX/galaxytools/tree/master/tools/msp_merge"
+homepage_url: "https://github.com/RECETOX/galaxytools"
+categories:
+  - Metabolomics
+repositories:
+  msp_merge:
+    description: "Utility tool to merge `msp` individual spectrum files into a collection of spectra in a single file."
+
```
hechth, line 9: remove blank line

tools/msp_merge/.shed.yml
```diff
@@ -0,0 +1,13 @@
+owner: recetox
+remote_repository_url: "https://github.com/RECETOX/galaxytools/tree/master/tools/msp_merge"
+homepage_url: "https://github.com/RECETOX/galaxytools"
+categories:
+  - Metabolomics
+repositories:
+  msp_merge:
+    description: "Utility tool to merge `msp` individual spectrum files into a collection of spectra in a single file."
```
hechth, line 8:
```suggestion
    description: "Utility tool to merge  individual `msp` spectra files in a collection into a single file."
```

tools/msp_merge/mergeMSP.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def merge_spectra(filenames, outfilename):
+    """Save MSP file containing merged spectra.
+
+    Args:
+        filenames   (str): Paths to MSP files from which to load each spectrum.
+        outfilename (str): Path to MSP file with merged spectra.
+    """    
+    return write_spectra(filenames, outfilename)
```
hechth, line 41: I think this one function which only forwards to another function at the moment can be removed and the `read_spectra(...)` and `save_as_msp(...)` calls can be in main directly.

tools/msp_merge/mergeMSP.xml
```diff
@@ -0,0 +1,49 @@
...
+        <organization
+            url="https://www.recetox.muni.cz/"
+            email="GalaxyToolsDevelopmentandDeployment@space.muni.cz"
+            name="RECETOX MUNI"/>
+    </creator>
+    <description>Merge MSP Spectra files</description>
+    <requirements>
+         <requirement type="package" version="0.14.0">matchms</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
```
hechth, line 22: Detect errors could be changed to aggressive.

tools/msp_merge/mergeMSP.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+listarg = argparse.ArgumentParser()
+listarg.add_argument('--filenames', nargs='+', type=str) 
+listarg.add_argument('--outfilename', type=str) 
+args = listarg.parse_args()
+outfilename = args.outfilename
+filenames = args.filenames
```
hechth, line 49: Could be checked for validity or left out since it is only picking up the names from args.

tools/msp_merge/mergeMSP.xml
```diff
@@ -0,0 +1,49 @@
...
+    <inputs>
+        <param name="input" type="data" format="msp" multiple="true"/>
+    </inputs>
+    <outputs>
+        <data name="outfile" format="msp" />
+    </outputs>
+    <tests>
+         <test>
+            <param name="input">
+              <collection type="paired">
```
hechth, line 37: Sure that this shouldn't be type `list`?

hechth (commented):
Please fix the minor issues, otherwise it looks good to me!

tools/msp_merge/mergeMSP.xml
maximskorik: @hechth, maybe `paired` is justified since the elements of this collection are static? Also, does it make any practical difference in this case?

tools/msp_merge/mergeMSP.xml
```diff
@@ -0,0 +1,49 @@
...
+                <element name="input" value="1-NITROPYRENE.msp"/>
+                <element name="input" value="2,3-DICHLOROPHENOL.msp"/>
+                <element name="input" value="2,4-DICHLOROPHENOL.msp"/>
+              </collection>
+            </param>
+            <output name="outfile" file="sample_output.msp" ftype="msp" />
+        </test>
+    </tests> 
+    <help><![CDATA[
+        Mergemsp handles input/output files in MSP format using the matchms.importing and matchms.exporting modules (https://matchms.readthedocs.io/en/latest/index.html).
```
maximskorik, line 47:
```suggestion
        MergeMSP handles input/output files in MSP format using the matchms.importing and matchms.exporting modules (https://matchms.readthedocs.io/en/latest/index.html).
```

maximskorik (approved):
Check out the suggestions when you have time. Otherwise looks good!

tools/msp_merge/mergeMSP.xml
wverastegui: @hechth and @maximskorik, both cases show same behavior.

tools/msp_merge/mergeMSP.xml
wverastegui: Done

tools/msp_merge/.shed.yml
wverastegui: Done (b6e44c7)

tools/msp_merge/mergeMSP.py
wverastegui: Changes applied ([4175a92](https://github.com/wverastegui/galaxytools/commit/4175a928f60307f3718ef64807d12135e070dcf4))

hechth:
All mentionings of the tool have to be the same. In the shed.yml it is called `msp_merge`, in the Galaxy wrapper the tool is is `mergmsp` and the xml file is called `mergeMSP.xml` - I think all of those have to have the same name for it to work.

tools/msp_merge/mergeMSP.xml
maximskorik: Though at least conventionally it should be the `list` type. See [here](https://planemo.readthedocs.io/en/latest/writing_advanced.html#id2).

tools/msp_merge/msp_merge.xml
```diff
@@ -0,0 +1,49 @@
...
+         <requirement type="package" version="0.14.0">matchms</requirement>
+    </requirements>
+    <command detect_errors="aggressive"><![CDATA[
+       #set newinput = str("' '").join([str($f) for $f in $input])
+        python3  $__tool_directory__/msp_merge.py
+        --filenames    '$newinput'
+        --outfilename  '$outfile'
+    ]]></command>
+    <inputs>
+        <param name="input" type="data" format="msp" multiple="true"/>
```
maximskorik, line 29: Planemo test likely fails because `type="data"` with `multiple="true"` attributes disallow to pass collection to this parameter at test time. See [here](https://planemo.readthedocs.io/en/latest/writing_advanced.html#processing-lists-reductions ).

maximskorik (changes requested):

tools/msp_merge/msp_merge.xml
```diff
@@ -0,0 +1,49 @@
...
+    </inputs>
+    <outputs>
+        <data name="outfile" format="msp" />
+    </outputs>
+    <tests>
+         <test>
+            <param name="input">
+              <collection type="paired">
```
martenson, line 37: `paired` is specific collection of only two elements with special relationships used in genomics, you probably want `list`

tools/msp_merge/msp_merge.xml
```diff
@@ -0,0 +1,53 @@
...
+    ]]></command>
+    <inputs>
+        <param name="filen" type="data" format="msp" multiple="true"/>
+    </inputs>
+    <outputs>
+        <data name="outfile" format="msp" />
+    </outputs>
+    <tests>
+         <test>
+            <param name="nfile" value="1-NITROPYRENE.msp,23-DICHLOROPHENOL.msp,24-DICHLOROPHENOL.msp" ftype="msp"/>
```
martenson, line 36: looking at inputs `nfile` looks like a typo

wverastegui: Second test (collection) does not pass

hechth:
@martenson I think now everything should be okay - only the collection test still causes some failure, but AFAIK that should not be the case? Or what are we missing here?

martenson:
I see both tests failing. 

```
Ran 2 tests in 75.910s

FAILED (errors=1, failures=1)
```


When trying this locally I see

```
Traceback (most recent call last):
  File "/Users/marten/devel/git/galaxytools/tools/msp_merge/msp_merge.py", line 30, in <module>
    save_as_msp(spectra, args.outfilename)
  File "/Users/marten/.local/lib/python3.9/site-packages/matchms/exporting/save_as_msp.py", line 39, in save_as_msp
    assert filename.endswith('.msp'), "File extension must be 'msp'."
AssertionError: File extension must be 'msp'.
```

martenson:
As of the collection, looking at the [docs](https://planemo.readthedocs.io/en/latest/writing_advanced.html#processing-lists-reductions) I think we missed an important difference:

>A noteworthy difference is that if a parameter of type data with multiple="true" is used, the elements of the collection are passed to the tool as a (Python) list, i.e. it is not possible:
> - to find out if a collection was passed,
> - to access properties of the collection (name,…), or
> - to write tests that pass a collection to the parameter (which would allow to name the elements explicitly).

tools/msp_merge/msp_merge.xml
hechth: Addressed in https://github.com/RECETOX/galaxytools/pull/263/commits/b7645dedc493c20f05ae92a170d0ff47e922e81c

tools/msp_merge/msp_merge.xml
hechth: Also fixed in https://github.com/RECETOX/galaxytools/pull/263/commits/b7645dedc493c20f05ae92a170d0ff47e922e81c

hechth:
> 

You're running an older version of matchms it seems - the capability to write out non `.msp` file ending files from matchms was added by @xtrojak quite some time ago.

hechth:
@wverastegui as mentioned by marten above, seems like using a collection in test cases is currently not possible - so let's rely on the first test case only.

martenson:
@hechth this is planemo running the test and it installed matchms@14.0 as requested in the recipe. Which change of matchms by xtrojak are you referring to?

tools/msp_merge/msp_merge.xml
wverastegui: Commas have been deleted. ([af48baa](https://github.com/RECETOX/galaxytools/pull/263/commits/af48baa577688acd2b5affa2ce15cd3aec5192eb))

hechth:
@martenson I think this looks good now - we can merge IMHO. How do you see it?

Issue #132, opened 2021-06-11
Tool to extract metadata table from sample .mzml and .json metadata files


hechth:
For RAMClustR & WaveICA, we need a metadata table with the structure shown below:

sample_name | class | sampleType | injectionOrder | batch
-- | -- | -- | -- | --
VT_160120_002 | sample | sample | 1 | 1
VT_160120_004 | sample | sample | 2 | 1
VT_160120_006 | sample | sample | 3 | 1
VT_160120_008 | sample | sample | 4 | 1
VT_160120_010 | sample | sample | 5 | 1

Sample type might be blank or QC.

This information should be extracted from the filename tag contained in the `.mzml` file and the metadata contained in the `.json` file.

martenson:
For testing purposes we have a sample metadata file corresponding to the GC seminal plasma dataset here: https://umsa.cerit-sc.cz/library/list#folders/F1c84aa7fc4490e6d/datasets/fdeac135082b95d1

maximskorik:
The batch information is also required for two-step-hybrid apLCMS: https://github.com/RECETOX/galaxytools/issues/174.

hechth:
This tool basically does what we need: https://github.com/ISA-tools/mzml2isa

bgruening:
A start for a Galaxy integration is here: https://github.com/ISA-tools/mzml2isa-galaxy

ping @tomnl

Tomnl:
Hi,

I have updated the tool to work Python 3

Let me know if you have questions regarding what information you need to extract from the mzML file

hechth:
Thank you very much :)

hechth:
@Tomnl I tested the tool and I'm getting the following error:

```
usage: mzml2isa -i IN_PATH -o OUT_PATH -s STUDY_ID [options]
mzml2isa: error: argument -s: expected one argument
```

I used the mzml collection as input data - any ideas?

PR #264, opened 2022-05-09
Add apLCMS Two-Step Hybrid wrapper

maximskorik:
* added apLCMS Two-Step Hybrid wrapper 
* added one copy of `.mzml` test-data input for multi-batch test
* added `.csv` test metadata
* bumped tool version to 0.9.5
---
### Roadmap:

- [x] merge https://github.com/RECETOX/recetox-aplcms/pull/34
- [x] update conda package to v0.9.4 https://github.com/bioconda/bioconda-recipes/pull/34825


tools/recetox_aplcms/recetox_aplcms_hybrid.xml
```diff
@@ -59,16 +59,26 @@
...
-        <expand macro="unsupervised_outputs"/>
-        <data name="updated_known_table" format="parquet" label="${tool.name} updated_known_table on ${on_string}"/>
+        <expand macro="unsupervised_outputs">
+            <collection  name="extracted_features" type="list" label="${tool.name} extracted_features on ${on_string}">
+                <discover_datasets pattern="__designation__" directory="extracted" format="parquet" />
+            </collection >
+            <collection  name="corrected_features" type="list" label="${tool.name} corrected_features on ${on_string}">
+                <discover_datasets pattern="__designation__" directory="corrected" format="parquet" />
+            </collection >
+        </expand>
```
xtrojak, line 81: Since the two output collections were removed from `unsupervised_outputs` macro, I think they should be reintroduced to `unsupervised.xml` outputs too. Or are they omitted for a reason?

xtrojak (changes requested):

tools/recetox_aplcms/macros.xml
```diff
@@ -146,6 +145,25 @@
                    help="The minimum number of raw data points to be considered as a true feature." />
         </section>
     </xml>
+    <xml name="multibatch_processing">
+        <section name="multibatch_processing" title="Multibatch processing">
+            <param name="min_within_batch_prop_detect" type="float" min="0" max="1" value="0.1"
+                   label="minimum_batchwise_prop_detect"
+                   help="The minimum proportion of samples in a batch that must have a given feature. The features that are less abundant than the value won't be extracted." />
```
hechth, line 152: The term `abundant` here is confusing. I'd suggest
`The minimum detection frequency of a feature for it to be included in the aligned feature table`.

tools/recetox_aplcms/macros.xml
```diff
@@ -146,6 +145,25 @@
...
         </section>
     </xml>
+    <xml name="multibatch_processing">
+        <section name="multibatch_processing" title="Multibatch processing">
+            <param name="min_within_batch_prop_detect" type="float" min="0" max="1" value="0.1"
+                   label="minimum_batchwise_prop_detect"
+                   help="The minimum proportion of samples in a batch that must have a given feature. The features that are less abundant than the value won't be extracted." />
+            <param name="min_within_batch_prop_report" type="float" min="0" max="1" value="0.5"
+                   label="minimum_batchwise_prop_report"
+                   help="The minimum proportion of samples in a batch that must have a given feature. The features that are less abundant than the value won't be reported in the final output." />
```
hechth, line 155: What does `reported` actually mean in this case? I think this is rather confusing - having a combined parameter with the detection frequency to extract the feature might make more sense, since the distinction between what is included in the table and what is reported is a bit confusing.

tools/recetox_aplcms/macros.xml
```diff
@@ -146,6 +145,25 @@
...
+            <param name="min_batch_prop" type="float" min="0" max="1" value="0.5"
+                   label="minimum_batch_prop"
+                   help="The minimum proportion of batches that must have a given feature. The features that are less abundant than the value won't be reported." />
+            <param name="batch_align_mz_tol" type="float" min="0" value="0.00001"
+                   label="batch_align_mz_tol"
+                   help="The m/z tolerance level for peak alignment within batch." />
+            <param name="batch_align_chr_tol" type="float" min="0" value="50.0"
+                   label="batch_align_chr_tol"
+                   help="The retention time tolerance level for peak alignment within batch." />
```
hechth, line 164: Are there extra parameters for alignment across batches or do they use the same values?

tools/recetox_aplcms/recetox_aplcms_hybrid.xml
hechth: I think so too - but seems they have been re-added.

tools/recetox_aplcms/recetox_aplcms_two_step_hybrid.xml
```diff
@@ -0,0 +1,111 @@
...
+                use.observed.range = $weak_signal_recovery.use_observed_range,
+                recover.min.count = $weak_signal_recovery.recover_min_count,
+                match.tol.ppm = $history_db.match_tol_ppm,
+                new.feature.min.count = $history_db.new_feature_min_count
+            )
+        ]]></configfile>
+    </configfiles>
+
+    <expand macro="inputs">
+        <param name="metadata" type="data" format="csv" label="metadata" help="A csv table with two columns, where the first column contains sample names and the second batch labels for each sample."/>
```
hechth, line 71: Is this table aligned with what WaveICA and RAMClustR expect as metadata table? I think it would be beneficial to just have a single table type for metadata.

tools/recetox_aplcms/recetox_aplcms_unsupervised.xml
```diff
@@ -53,7 +53,11 @@
     </configfiles>
 
     <expand macro="inputs">
-        <expand macro="noise_filtering"/>
+        <expand macro="noise_filtering">
+            <param name="min_exp" type="integer" min="1" value="2"
+            label="min_exp"
+            help="If a feature is to be included in the final feature table, it must be present in at least this number of spectra." />
```
hechth, line 59:
```suggestion
            help="If a feature is to be included in the final feature table, it must be present in at least this number of samples." />

```

hechth (commented):
Seems fine, just some minor comments which could be improved. I'll test the Tool later today to check the runtime etc. :)

hechth:
Also one of the files is larger than 1MB - how can we deal with this somehow?
@martenson any ideas?

maximskorik:
> Also one of the files is larger than 1MB - how can we deal with this somehow? @martenson any ideas?

Also, this new file is the exact copy of `mbr_test0.mzml`. Maybe that's possible to make two different symlinks for the same file during test-time?

tools/recetox_aplcms/recetox_aplcms_hybrid.xml
maximskorik: Great point, @xtrojak! I totally forgot about `unsupervised.xml` on this occasion. Fixed in 1a1e6a0.

tools/recetox_aplcms/macros.xml
maximskorik: You're right. It does make more sense to have one parameter for that. Especially because when `min_within_batch_prop_detect` is lower than `min_within_batch_prop_report` (as by default) only the former has practical significance. Removed `min_within_batch_prop_report` in 3e752f9.

tools/recetox_aplcms/macros.xml
maximskorik: Changed description in 3e752f9.

tools/recetox_aplcms/macros.xml
maximskorik: The Two-Step Hybrid has distinct parameters for aligning _m/z_ and _rt_ between batches and within batches.

tools/recetox_aplcms/recetox_aplcms_two_step_hybrid.xml
maximskorik: Yes, the tool will work with the metadata table described here: https://github.com/RECETOX/galaxytools/issues/132

tools/recetox_aplcms/recetox_aplcms_unsupervised.xml
maximskorik: Corrected in 5ba2980. 
Also, `min_exp` parameter, used in **Unsupervised** and **Hybrid** tools, is essentially an absolute form of  `min_within_batch_prop_detect` form **Two-Step Hybrid**. It may make more sense to have all tools operate with a relative parameter instead of the absolute.

martenson:
I think you can straight up use symlinks, something like `ln -s mbr_test0.mzml mbr_test4.mzml` should do the job.

hechth:
> yeah, I think you can just do this: [36e0539](https://github.com/RECETOX/galaxytools/commit/36e053928443e2e9599ff5902e6483739b7278ec)
> 
> As of the CI check, I wouldn't remove it, just ignore it in cases you think it is safe to do.
> 
> p.s. this PR should be squashed before merging, to actually not use the extra space

Good idea with the squashing - @maximskorik do you know how to do this?

hechth:
I think the symlink actually has to be renamed to be the same as the file before (`mbr_test0_copy.mzml`), otherwise the tests won't pass - the filenames are I think part of the output.

maximskorik:
@martenson, what's the difference between the two tests that the **Test tools** action executes? I'm not sure why one of them fails.



> Good idea with the squashing - @maximskorik do you know how to do this?

@hechth, selecting **Squash and merge** when merging should be it, right?

hechth:
@maximskorik I think this is one option, then it will all be one commit - you can actually also squash only certain commits via the commandline.

martenson:
> What's the difference between the two tests that the Test tools action executes? I'm not sure why one of them fails.

Could you elaborate here? I do not understand.

maximskorik:
> Could you elaborate here? I do not understand.

Sorry, I didn't pay enough attention. One out of the two tests in **Galaxy Tool Linting and Tests for push and PR / Test tools (0, 3.7)** failed. I didn't realize these were the tests for `recetox_aplcms_hybrid` and not `recetox_aplcms_two_step_hybrid`, so I couldn't figure out where the second test came from.

maximskorik:
> Sorry, I didn't pay enough attention. One out of the two tests in Galaxy Tool Linting and Tests for push and PR / Test tools (0, 3.7) failed. I didn't realize these were the tests for recetox_aplcms_hybrid and not recetox_aplcms_two_step_hybrid, so I couldn't figure out where the second test came from.

One of the `recetox_aplcms_hybrid` tests fails because of rounding. When comparing the actual output with the expected via `testthat::expect_equal()`, the two outputs are the same. So I think we could ignore that or change the expected data Though, I think changing the expected data every time we adjust the tool is not an error-prone approach. Maybe it's better to just double-check equality with `testthat::expect_equal()` from now on.

The same issue occurs with `recetox_aplcms_unsupervised` tests. Both table pairs are actually equal when compared via `testthat::expect_equal()`.

@martenson, @hechth, what you think: do we ignore that or change the expected data?

martenson:
@maximskorik Why do you think is the rounding different? Is it a difference between running inside/outside container?

I'd probably ignore the rounding error in the test with something like `lines_diff` but since we are working with a parquet file it is not as straightforward.

maximskorik:
> @maximskorik Why do you think is the rounding different? Is it a difference between running inside/outside container?

I suspect it's caused by the conversion between **R** data-frames and **dplyr** data-frames. This may be the reason why the `recetox_aplcms_hybrid` [test](https://github.com/RECETOX/galaxytools/blob/5de7a58a9cb41de93e13d3acb211cf1058f367d0/tools/recetox_aplcms/recetox_aplcms_hybrid.xml#L99) for data in recetox format does not fail – the expected data was converted to **dplyr** prior to saving.

maximskorik:
> My understanding is that within the Galaxy tool the rounding will be consistent -- the inconsistency is when running R tests directly. Is this correct?

@martenson Yes, I think that's correct. At least that's what I've observed so far.

I've updated the test data in e92f0d6 and made sure that they match the old data via R's `testthat::expect_equal(..., tolerance = 1e-9)` (this tolerance is just an arbitrarily low number for some reference; lower values also passed).

martenson (approved):

tools/recetox_aplcms/recetox_aplcms_hybrid.xml
```diff
@@ -59,16 +59,26 @@
...
```
hechth, line 81: Is this a way of combining a macro and other parameters in one block?

hechth (approved):
Looks good - nice work!

tools/recetox_aplcms/recetox_aplcms_hybrid.xml
maximskorik: Yes, you can expand your macros like this by placing `<yield />` where you need some parametrization. Here's another example from Planemo [docs](https://planemo.readthedocs.io/en/latest/writing_advanced.html#parameterizing-xml-macros-with-yield).

PR #269, opened 2022-06-14
apLCMS: input collections

### LINKED ISSUE #268: apLCMS: incorrect input collections handling
The inputs in `apLCSM` given as collections of files are handled incorrectly. After reading [this](https://planemo.readthedocs.io/en/latest/writing_advanced.html#collections), they need to specified as `type="data_collection"` instead of `multiple="true" min="2"` and passed as a proper collection to in the tests.

xtrojak:
Passing the input collections was implemented [incorrectly](https://github.com/RECETOX/galaxytools/issues/268), which made it impossible to use the `apLCMS` steps in workflows. The new implementation in this PR should enable this. 

TODO: 
- only collections are allowed now, not possible to pass separate files as inputs
- execution fails outside of tests

Close #268


hechth:
@xtrojak with execution failing outside of tests you mean with `planemo serve` for example?

xtrojak:
> @xtrojak with execution failing outside of tests you mean with `planemo serve` for example?

Actually it works fine, it was my bad.

hechth:
I think it is fine to only pass collections - I mean you can pass a collection with a single file, which is fine.

hechth (approved):

martenson (approved):
works well, thanks for the adjustments

PR #281, opened 2022-06-30
matchms similarity: Added NeutralLossesCosine metric 

### LINKED ISSUE #278: matchms similarity: Add NeutralLossesCosine metric
This should be rather straightforward for the [NeutralLossesCosine](https://github.com/matchms/matchms/blob/master/matchms/similarity/NeutralLossesCosine.py) - currently has limited applications since we have little LC-MS2 data.

For a reference on how other similarity metrics are implemented inside a Galaxy wrapper see [here](https://github.com/RECETOX/galaxytools/blob/master/tools/matchms/matchms_similarity.xml).

Relevant literature can be found [here](https://www.researchgate.net/publication/361051029_Comparison_of_Cosine_Modified_Cosine_and_Neutral_Loss_Based_Spectral_Alignment_For_Discovery_of_Structurally_Related_Molecules).

zargham-ahmad:
Added NeutralLossesCosine metric and test data to matchms similarity which will close #278.


tools/matchms/macros.xml
```diff
@@ -1,5 +1,5 @@
 <macros>
-    <token name="@TOOL_VERSION@">0.14.0</token>
+    <token name="@TOOL_VERSION@">0.15.0</token>
```
hechth, line 2: Is the neutral losses cosine added in 0.15.0? If so, please state the version bump in the PR description

hechth (commented):
Looks fine - the documentation could get an update with the reference to the paper describing the losses and there could be a small section explaining the metrics.

If it is possible to add a help text to options that would be great - otherwise the metrics could be explained a bit in the footnote text or so.

hechth:
Since the tool version increased, the Galaxy version should be set back to 0 in all tools.

Otherwise looks good to me and ready for review.

maximskorik:
@hechth, I guess @zargham-ahmad can add himself to creators as well?

hechth:
@maximskorik True! I guess this also applies to you for the new metrics.
@zargham-ahmad Please add yourself under the creator section.

hechth (approved):

hechth:
@zargham-ahmad Looks good to me, but please make sure that the linting actually passes to keep the tests green :).

zargham-ahmad:
> @maximskorik True! I guess this also applies to you for the new metrics. @zargham-ahmad Please add yourself under the creator section.

added myself in creator in cb7d2ec
I don't know why one of the test in the [Test tools (0, 3.7)](https://github.com/RECETOX/galaxytools/runs/7280971468?check_suite_focus=true#logs) failed in the last commit because the error doesn't make sense and now it passed.
```
File "/home/runner/.planemo/planemo_tmp_ncp_ljmy/matchms_filtering_wrapper.py", line 4, in <module>
    from matchms.exporting import save_as_mgf, save_as_msp
ModuleNotFoundError: No module named 'matchms'
```

hechth:
Also no idea, but if it passes I think we can merge it.

PR #289, opened 2022-07-19
matchms: switch to `Scores` objects

maximskorik:
### Changes
- changed **matchms similarity** output to `JSON` file containing serialized `Scores` object
- changed  **matchms formatter** input to `JSON` file containing serialized `Scores` object
- updated the test data to respect these changes
### Testing

To ensure that new test data are equivalent to the old ones the following tests have been done:
- the new expected `JSON` outputs from **matchms similarity** were imported, their scores transformed into pandas `DataFrame` and compared to `DataFrames` produced from old `tsv` data with `pandas.testing.assert_frame_equal(...)`
- **matchms formatter** were run on the new `JSON` input, and the output was compared to expected data with `pandas.testing.assert_frame_equal(...)`

### TODO

- [ ] bump matchms `@TOOL_VERSION@` when https://github.com/matchms/matchms/pull/353 is released


maximskorik:
The CI check fails because the test files for score objects are rather large (up to 70Mbs). @ hecht, do you have any preferences about whether we ignore that or limit ourselves to testing the wrapper on smaller MSP datasets (hence smaller outputs)?

hechth (approved):

hechth:
@maximskorik Looks good - only thing is the file sizes. Is it somehow possible to compress the `json` files to be more compact, something like zip or so? Or maybe we can make the test files a little bit smaller so that the scores objects are actually smaller.

maximskorik:
**Combine chunked test results** fails both here and in #286. I'm not sure what's going on. Is there a way for me to check the workflow artifacts?

xtrojak:
> **Combine chunked test results** fails both here and in #286. I'm not sure what's going on. @hechth Is there a way for me to check the workflow artifacts?

@maximskorik you can access workflow artifacts in the Summary of the workflow, scroll down for artifacts. In this case, the error is

```
Job in error state.. tool_id: matchms_formatter, exit_code: 0, stderr: 
Matplotlib created a temporary config/cache directory at /tmp/matplotlib-yllesa_q 
because the default path (/home/runner/.config/matplotlib) is not a writable directory; 
it is highly recommended to set the MPLCONFIGDIR environment variable to 
a writable directory, in particular to speed up the import of Matplotlib and to
better support multiprocessing.
Fontconfig error: No writable cache directories
```

maximskorik:
So I've managed to reduce the `stderr` to:
```
Job in error state.. tool_id: matchms_formatter
exit_code: 0
stderr: Fontconfig error: No writable cache directories
```

I haven't yet managed to find any hint or solution as to where this `Fontconfig error` comes from and how to fix that.

PR #314, opened 2022-12-06
add wrapper for training Spec2Vec in Galaxy

### LINKED ISSUE #315: add wrapper for Spec2Vec model training

maximskorik:
### Updates
- add `spec2vec_training` tool to Galaxy

### Description
This tool can train a Spec2Vec model from `msp` or `mgf` spectra files. The outputs are `json` model-metadata and `npy` peak-embeddings, which can be read by `spec2vec.serialization` module (in the future downstream step) to compute Spec2Vec similarity scores. Optional outputs are Python `pickle` file & checkpoints of a model at user-defined iterations.

Closes #315


xtrojak:
@maximskorik can you please explain the role of docker image? Is it just to get the newest version available in `master` branch?

maximskorik:
> @maximskorik can you please explain the role of docker image? Is it just to get the newest version of the tool available in `master` branch?

Exactly. We recently added functionality for Spec2Vec to export a model to `json`+`npy` files and it hasn't been released yet. Hence the docker container which points to the commit with this update until these changes are released and become available via Conda.

tools/spec2vec/spec2vec_training.xml
```diff
@@ -0,0 +1,222 @@
...
+                <when value="FALSE">
+                    <param label="Number of training epochs" name="epochs" type="integer" value="10"
+                           help="Number of epochs to train the model."/>
+                </when>
+            </conditional>
+        </section>
+
+        <section title="Training hyperparameters" name="training_parameters" expanded="true">
+            <param label="Vector size" name="vector_size" type="integer" value="300"
+                   help="Dimensionality of the feature vectors (i.e., into how many dimensions to encode each m/z and neutral loss peak."/>
```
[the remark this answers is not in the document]
maximskorik: Yes, precisely! This parameter is the size of embeddings that represent peaks and losses in a latent space. I haven't thought of it from this point of view.

tools/spec2vec/spec2vec_training.xml
```diff
@@ -0,0 +1,222 @@
...
+            --ns_exponent $training_parameters.hs_param.ns_exponent \
+            #end if
+            --sorted_vocab $training_parameters.sorted_vocab \
+            --batch_words $training_parameters.batch_words \
+            --shrink_windows $training_parameters.shrink_windows \
+            #if $training_parameters.trim_vocab.max_vocab_size_bool
+            --max_vocab_size $training_parameters.trim_vocab.max_vocab_size \
+            #end if
+            --n_decimals $training_parameters.n_decimals \
+            --n_workers \${GALAXY_SLOTS:-1} \
```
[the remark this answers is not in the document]
xtrojak: _Short answer:_ I think so.

_Long answer:_ When `Slurm` executes a job, it can add `--cpus-per-task <num>` to the call, which can be configured using `Ansible` for the specific tool (using destinations). From what I have [read](https://galaxyproject.org/admin/config/galaxy-slots/), this number should then be available as environment variable `GALAXY_SLOTS`. It is being assigned in [this](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/util/job_script/CLUSTER_SLOTS_STATEMENT.sh) script.

maximskorik:
The CI fails because the linter is concerned about suffixes in the `value` and `delta` attributes of `<has_size>`. I've submitted an issue to Planemo https://github.com/galaxyproject/planemo/issues/1335.

xtrojak (approved):
Looks pretty good! Great job @maximskorik

tools/spec2vec/spec2vec_training.xml
```diff
@@ -0,0 +1,254 @@
...
+                   help="Add a Pickle output besides default JSON."/>
+            <conditional name="model_checkpoints">
+                <param label="Model checkpoints" name="is_true" type="boolean" checked="false" truevalue="TRUE" falsevalue="FALSE"
+                   help="Epochs after which to save a model."/>
+                <when value="TRUE">
+                    <param label="Number of training epochs with checkpoints" name="checkpoints" type="text" value="10,20,50"
+                           help="Comma-separated epoch numbers after which to save a model. The highest number will be used as a total number of epochs for training.">
+                           <validator type="empty_field"/>
+                           <validator type="regex" 
+                                message="The input has to be a comma-separated sequence of integers without trailing commas. For example: 10,20,50">^[0-9][0-9,]*[0-9]$</validator>
```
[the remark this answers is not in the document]
hechth: Yeah - regex can be tricky sometimes :D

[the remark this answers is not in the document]
maximskorik: fixed in f55bc72
cc @hechth

hechth:
Can we maybe change this to be compliant with the CI so that it passes? I know it is not a fault on our side, but still as soon as the CI fails and you merge stuff, it becomes harder to see when it actually starts failing for a different reason.

tools/spec2vec/spec2vec_training.xml
```diff
@@ -0,0 +1,254 @@
+<tool id="spec2vec_training" name="Spec2Vec Model Training" version="@TOOL_VERSION@-@TOOL_DEV_VERSION@+galaxy0" python_template_version="3.5" profile="21.05">
+    <description>Train a Spec2Vec model for mass spectra similarity scoring</description>
+
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="creator"/>
+
+    <requirements>
+        <container type="docker">recetox/spec2vec:@COMMIT_SHA@</container>
```
bgruening, line 10: I have created a conda package here: https://github.com/bioconda/bioconda-recipes/pull/38439

This will come automatically with a container. Would it be possible to use the package definition here?

tools/spec2vec/spec2vec_training.xml
```diff
@@ -0,0 +1,254 @@
...
+        sh ${weights_symlink} &&
+        sh ${spec2vec_python_cli}
+    ]]></command>
+        
+    <configfiles>
+        <configfile name="weights_symlink">
+            ln -fs '${weights_filename}' '${weights_filename}.npy'
+        </configfile>
+        <configfile name="spec2vec_python_cli">
+            python3 ${__tool_directory__}/spec2vec_training_wrapper.py \
```
bgruening, line 23: single quotes are missing here

tools/spec2vec/spec2vec_training.xml
```diff
@@ -0,0 +1,254 @@
...
+    </requirements>
+
+    <command detect_errors="exit_code"><![CDATA[
+        sh ${weights_symlink} &&
+        sh ${spec2vec_python_cli}
+    ]]></command>
+        
+    <configfiles>
+        <configfile name="weights_symlink">
+            ln -fs '${weights_filename}' '${weights_filename}.npy'
```
bgruening, line 20: Is there any reason why you use a configfile here and not include this directly in the command section?

tools/spec2vec/spec2vec_training.xml
```diff
@@ -0,0 +1,254 @@
...
+
+    <inputs>
+        <param label="Training spectra" name="spectra_filename" type="data" format="msp,mgf"
+               help="Spectra file to train a Spec2Vec model."/>
+
+        <section title="Output parameters" name="output_parameters" expanded="true">
+            <param label="Save model as Pickle file" name="as_pickle" type="boolean" checked="false" truevalue="TRUE" falsevalue="FALSE"
+                   help="Add a Pickle output besides default JSON."/>
+            <conditional name="model_checkpoints">
+                <param label="Model checkpoints" name="is_true" type="boolean" checked="false" truevalue="TRUE" falsevalue="FALSE"
```
bgruening, line 71: The IUC guidelines recommend using a select field here: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#booleans

tools/spec2vec/spec2vec_training.xml
```diff
@@ -0,0 +1,254 @@
...
+            <param label="Shrink windows" name="shrink_windows" type="boolean" checked="true" truevalue="TRUE" falsevalue="FALSE"
+                   help="EXPERIMENTAL. If true, the effective window size is uniformly sampled in range [1,Window] for each target peak during training."/>
+            <conditional name="trim_vocab">
+                <param label="Maximum unique peaks and neutral losses in the spectral vocabulary" name="max_vocab_size_bool" type="boolean" truevalue="TRUE" falsevalue="FALSE" checked="false"
+                        help="Limits the RAM during vocabulary building; if there are more unique peaks and neutral losses than this, then prune the infrequent ones. Disable for no limit (default)."/>
+                <when value="TRUE">
+                    <param label="Maximum unique peaks and neutral losses" name="max_vocab_size" type="integer" value="100000"/>
+                </when>
+            </conditional>
+            <param label="Number of decimals to round m/z values" name="n_decimals" type="integer" value="2"
```
bgruening, line 149: You could add min/max to those paramters.

tools/spec2vec/spec2vec_training_wrapper.py
```diff
@@ -0,0 +1,108 @@
+import argparse
```
bgruening, line 1: A shebang would be nice here `#!/usr/bin/env python`

tools/spec2vec/spec2vec_training.xml
bgruening: This entire configfile is only a call to a python script that is on its own is call to a simple function.

If the python script was just created for this Galaxy tool you can avoid the argparse and the python script all together, you can put the function call directly into the `configfile` here. You essentially create the python script on the fly via Galaxy and execute it.

tools/spec2vec/spec2vec_training.xml
hechth: We actually discussed this with @florian-huber - the dockerfile allows us to test development builds on galaxy so to say, since we can always build and upload a docker container and then for the correct versions we plan to use the conda package - so this comes in handy!

tools/spec2vec/spec2vec_training.xml
hechth: I personally like the separation of having the individual wrapper file that handles the arguments and stuff on the python side related to the package and then having the galaxy related stuff in the tool script. I think the command line parsing might also at some point migrate into the spec2vec package to make that more user friendly, and then we can use it from here.

tools/spec2vec/spec2vec_training.xml
hechth: That is indeed a good idea!

tools/spec2vec/spec2vec_training.xml
bgruening: Yes, I agree. If this is a separate package it is the preferred way.

tools/spec2vec/spec2vec_training.xml
maximskorik: It appears to me that storing symlinks in a config file would improve code readability.

maximskorik: After more thoughts on this, using one oneliner instead of another oneliner doesn't seem like much of a readability improvement. Moved symlinks to command in 4e0fab4.

tools/spec2vec/spec2vec_training.xml
maximskorik: Thank you for the suggestion! I wasn't aware of that. Switched the input to `select` in 0cc4fa0.

tools/spec2vec/spec2vec_training.xml
maximskorik: Added more validators in 662d691.

tools/spec2vec/spec2vec_training_wrapper.py
maximskorik: Fixed in 8ec040a.

tools/spec2vec/spec2vec_training.xml
bgruening: For small code like this it also helps admins later I assume. Now the command ends up in the main tool_script.sh file and it should be obvious what's happening for an Galaxy admin. With a config file this is partially hidden in an additional file.

maximskorik: That's a good point!

tools/spec2vec/spec2vec_training.xml
bgruening: You are very welcome! Thanks for all your effort in those cool tools!

Issue #291, opened 2022-07-29
recetox-aplcms-extract-features input form


hechth:
- [x] `sigma_ratio_lim_min` & `sigma_ratio_lim_max` should be conditional on bi-Gaussian peak shape
- [x] ~~`moment_power` is likely only applicable if estimation mode is moment~~
- [ ] explain `BIC_factor` better
- [x] ~~`make `baseline_correct_noise_percentile` only appear if the `baseline_correct` is NA`~~
- [ ] explain bandwidth parameters better
- [ ] explain `sd_cut_min` and `sd_cut_max` parameters better

xtrojak:
Actually `baseline_correct_noise_percentile` is used only in plotting of histograms, so we can remove it.

PR #317, opened 2022-12-12
recetox-aplcms `v0.10.1`

### LINKED ISSUE #169: aplcms: replace galaxy tests with real data
The aplcms tool tests rely on somewhat artificial data. This could be replaced by a small but realistic dataset.

### LINKED ISSUE #184: aplcms: Improve documentation
- [ ] Add documentation regarding which output tables are what and which columns they contain etc. using examples.
- [x] Improve the documentation on the parameters
- [ ] Mark optional parameters as optional -> deduce the information from the paper and help text
- [ ] Explain `rt`, `metadata`, `intensity` tables

### LINKED ISSUE #308: apLCMS: update wrappers to `v0.10.0`
With the new release `v0.10.0` of `apLCMS`, the wrappers need to get updated. Most of them will be completely replaced.

Wrappers that need to be introduced:

- [ ] `remove_noise` - call `proc.cdf`
- [ ] `generate_feature_table` - call `prof.to.features`
- [ ] `compute_clusters`
- [ ] `compute_template`
- [ ] `correct_time`
- [ ] `merge_known_table` - call `merge_features_and_known_table`

Wrappers to be changed:

- [ ] `align_features` - call `create_aligned_feature_table` instead of `align_features`
- [ ] `recover_weaker_signals` - executed in parallel per corrected features table

Obsolete wrappers:

- [ ] `adjust_time`
- [ ] `extract_features`
- [ ] `hybrid`
- [ ] `unsupervised`
- [ ] `two_step_hybrid`

The amount of code in `utils.R` and `main.R` should also be significantly reduced. Wrappers for `unsupervised` and `hybrid` are obsolete because they can be replaced by workflows. Special attention needs to be paid to the outputs of the workflows, as it is not only output of the last step.

xtrojak:
Preparation of wrappers for the new version `v0.10.1` of `recetox-aplcms`.
Details on introduced and changed wrapper are available in #308.

We have introduced an approach in the wrappers that stores the attribute `sample_name` in the `.parquet` files. This way, we don't have to rely on filenames but use sample names to actually sort and keep the feature tables in the correct order. This is example code how it can be done (but is already implemented in the wrappers):

```R
table <- arrow::read_parquet('file.parquet')
attributes(table) # no sample_name present
attr(table, "sample_name") <- "theSampleXYZ"
attributes(table) # now it contains sample_name
arrow::write_parquet(table, "output.parquet")

loaded_table <- arrow::read_parquet('output.parquet')
attributes(loaded_table) # still contains sample_name
attr(loaded_table, "filename") # this is how it can be accessed
```

TODO list:

- [x] add all newly introduced wrappers, defined their inputs, respective `recetox-aplcms` function call, and outputs.
- [x] remove obsolete wrappers
- [x] clean up macro file
- [x] update utils R functions and remove the obsolete ones
- [x] actually run and test the wrappers once the bioconda package is available, make sure all runs smoothly and keep in mind the individual wrappers need to "fit together" to form 'unsupervised' and 'hybrid' workflows
- [x] solve how to get value of `sample_name` from input `mzml` files (this concerns `remove_noise` and `recover_weaker` wrappers) - do we want to use filename? Are these values present inside the files? (see 28d28db)
- [x] finish `select_table_with_sample_name` (in `utils.R`) function - make sure it works (see https://github.com/RECETOX/galaxytools/pull/317#discussion_r1054386921)
- [ ] we will not include tests in the individual wrappers - instead, we will write [workflow tests](https://github.com/RECETOX/workflow-testing/pull/14). For that we will reuse package [test data](https://gitlab.ics.muni.cz/umsa/umsa-files). The main motivation is that for meaningful tests, a larger dataset is required, and in workflows testing they can be downloaded from a remove location instead of storing them locally with the wrappers. This should also resolve #169.
- [x] so far, we are relying on the presence of the attribute `sample_name` in the `.parquet`, this needs to be handled properly in the case it is missing (raise a proper error?)
- [x] `compute_clusters` should have a switch to enter the input tolerances as numbers or use an input parquet file
- [ ] #291 
- [x] #184 

Close #308, closes #184


hechth:
>  so far, we are relying on the presence of the attribute sample_name in the .parquet, this needs to be handled properly in the case it is missing (raise a proper error? use the filename instead and hope for the best?)

There should be an exception raised in the wrapper when loading the data.

hechth:
> solve how to get value of sample_name from input mzml files (this concerns remove_noise and recover_weaker wrappers) - do we want to use filename? Are these values present inside the files?

This should be read from the file using the `mzR` package

maximskorik:
> > solve how to get value of sample_name from input mzml files (this concerns remove_noise and recover_weaker wrappers) - do we want to use filename? Are these values present inside the files?
> 
> This should be read from the file using the `mzR` package

After some research, it appears that the actual sample names are stored in `id` attribute of `run` subelement of `mzml` XML file (http://www.peptideatlas.org/tmp/mzML1.1.0.html#run). Unfortunately, `mzR` doesn't read sample names – its `fileName` method always returns the path used to read a file. So far, the only option to get that attribute is with `pymzml` package. Alternatively, it may be possible to parse XML with some R package.

tools/recetox_aplcms/recetox_aplcms_recover_weaker_signals.xml
```diff
@@ -1,143 +1,91 @@
...
         <configfile name="run_script"><![CDATA[
-            #set filenames = str("', '").join([str($f.element_identifier) for $f in $ms_files])
-            filenames <- sort_samples_by_acquisition_number(c('$filenames'))
+            sample_name <- Sys.getenv("SAMPLE_NAME", unset = NA)
 
-            #set extracted_files = str("', '").join(["extracted_" + str($f.element_identifier) for $f in $extracted_files])
-            extracted <- load_features(c('$extracted_files'))
+            #set extracted_filenames = str("', '").join([str($f.element_identifier) for $f in $files])
+            extracted_feature_tables <- load_parquet_collection(extracted_filenames)
+            extracted_features <- select_table_with_sample_name(extracted_feature_tables, sample_name)
```
maximskorik, line 19: This implementation has a problem. `load_parquet_collection` loads the parquet tables and converts them to matrices with `as.matrix`. After converting to matrices, the tables no longer store the attributes that the parquet files did. Hence, `select_table_with_sample_name` fails because sample names are no longer in the attributes. @xtrojak, is it necessary to convert the data frames to matrices?

maximskorik: Besides that, `select_table_with_sample_name` works as expected with parquet data frames.

xtrojak:
Bioconda package for `v0.10.0` was released today https://github.com/bioconda/bioconda-recipes/pull/38281.

xtrojak:
>

xtrojak:
> > > solve how to get value of sample_name from input mzml files (this concerns remove_noise and recover_weaker wrappers) - do we want to use filename? Are these values present inside the files?
> > 
> > 
> > This should be read from the file using the `mzR` package
> 
> After some research, it appears that the actual sample names are stored in `id` attribute of `run` subelement of `mzml` XML file (http://www.peptideatlas.org/tmp/mzML1.1.0.html#run). Unfortunately, `mzR` doesn't read sample names – its `fileName` method always returns the path used to read a file. So far, the only option to get that attribute is with `pymzml` package. Alternatively, it may be possible to parse XML with some R package or bash script (checking if that's feasible).
> 
> UPD: see [28d28db](https://github.com/RECETOX/galaxytools/commit/28d28dbf67338b2720d9d0f024ceec5929e61299). `grep` pattern: https://unix.stackexchange.com/a/529674/536940

Unfortunately, the `id` attribute of `run` subelement is not always present (for example, in our [RCX test data](https://gitlab.ics.muni.cz/umsa/umsa-files/-/tree/master/testdata/recetox-aplcms/input)). The question is whether this should lead to an error or if we want to use the filename somehow instead.

hechth:
@xtrojak the test files exported from mzmine are actually also not great examples regarding metadata as they actyally lose quite a lot of it during the export - but no idea how to shorten the data and make sure the exported metadata is maintained. We could modify our test files and re-include the metadata headers, but I mean everyone who will manipulate and then export their data from mzmine will run into the same issue.

xtrojak:
> @xtrojak the test files exported from mzmine are actually also not great examples regarding metadata as they actyally lose quite a lot of it during the export - but no idea how to shorten the data and make sure the exported metadata is maintained. We could modify our test files and re-include the metadata headers, but I mean everyone who will manipulate and then export their data from mzmine will run into the same issue.

We need to consider a general approach. Maybe we should just use the filenames, whatever they are, or take the files in the same order they are given in the collection. This could be a requirement for the user.

This concerns only the first step `remove_noise` (and also `recover_weaker` but the same solution can be used there).

bgruening:
As usual, I don't understand much here :( I can only recommend not relying on filenames if possible. If you mean with filenames the name you see in Galaxy at the dataset in the UI - those are just Galaxy metadata and those can be changed at any time even after the job has be finished. They also can contain emojis and stuff, that you don't want to have on your commandline or anywhere. Names (element identifiers) in collections are more stable, because collections are immutable, but then you restrict yourself to collections and users need to create meaningful element identifiers in the first place.

Hope that helps somehow.

xtrojak:
You are entirely right @bgruening, we were relying on filenames in the previous version, and it didn't work well.

We decided to use the `id` present inside the input `mzML` files and require them to be correct. For that, we can employ an [mzML validator](https://github.com/RECETOX/galaxytools/pull/319) as a first step. This ensures they are formatted according to mzML specification. However, two issues still need to be resolved: (1) the [regex](https://github.com/xtrojak/galaxytools/blob/d9d654abd7ebd7c4c0452eebf6195f1e89fff098/tools/recetox_aplcms/recetox_aplcms_remove_noise.xml#L10) to read the `id` expects it to be the first parameter of the `run` subelement and (2) `id` across samples need to be different.

maximskorik:
> You are entirely right @bgruening, we were relying on filenames in the previous version, and it didn't work well.
> 
> We decided to use the `id` present inside the input `mzML` files and require them to be correct. For that, we can employ an [mzML validator](https://github.com/RECETOX/galaxytools/pull/319) as a first step. This ensures they are formatted according to mzML specification. However, two issues still need to be resolved: (1) the [regex](https://github.com/xtrojak/galaxytools/blob/d9d654abd7ebd7c4c0452eebf6195f1e89fff098/tools/recetox_aplcms/recetox_aplcms_remove_noise.xml#L10) to read the `id` expects it to be the first attribute of the `run` subelement and (2) `id` across samples need to be different.

As to (1), we decided to use [`pymzml`](https://anaconda.org/bioconda/pymzml) instead of the regex to look up run `id` in mzMLs.

xtrojak:
> We decided to use the `id` present inside the input `mzML` files and require them to be correct. For that, we can employ an [mzML validator](https://github.com/RECETOX/galaxytools/pull/319) as a first step. This ensures they are formatted according to mzML specification. However, two issues still need to be resolved: (1) the [regex](https://github.com/xtrojak/galaxytools/blob/d9d654abd7ebd7c4c0452eebf6195f1e89fff098/tools/recetox_aplcms/recetox_aplcms_remove_noise.xml#L10) to read the `id` expects it to be the first attribute of the `run` subelement and (2) `id` across samples need to be different.

Issue (2) could be solved the following way: the wrappers working with individual files can write a warning when there is no `sample_name`, and wrappers using collections would check whether all files contain a unique non-`NA` `sample_name` (since these wrappers actually require the `sample_names` to be present).

tools/recetox_aplcms/recetox_aplcms_remove_noise.xml
```diff
@@ -0,0 +1,59 @@
+<tool id="recetox_aplcms_remove_noise" name="RECETOX apLCMS - remove noise" version="@TOOL_VERSION@+galaxy0">
+    <description>filter noise and detect peaks from LC/MS data</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="creator"/>
+    <expand macro="requirements">
+        <requirement type="package" version="2.5.2">pymzml</requirement>
```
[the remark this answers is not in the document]
maximskorik: Good point. I guess there is no need to have two separate containers for some steps of the same tool. If anything, I suspect it can create confusion if it installs different versions of some unpinned dependencies in different containers.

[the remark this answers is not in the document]
xtrojak: Unified in [8dae925](https://github.com/RECETOX/galaxytools/pull/317/commits/8dae9254ae5e02962cc504417d0cc81cf4730f6f).

tools/recetox_aplcms/mzml_id_getter.py
```diff
@@ -0,0 +1,25 @@
...
+    parser = argparse.ArgumentParser(description='Get run ID from an mzML file.')
+    parser.add_argument('mzml_file', help='Path to an mzML file to get run ID from.')
+    args = parser.parse_args()
+
+    mzml = Reader(args.mzml_file)
+    id = mzml.info['run_id']
+
+    if id is None:
+        sys.stderr.write('ERROR: The mzML file has no run ID.')
+        sys.exit(1)
```
xtrojak, line 19: The `remove_noise` wrapper will fail when the `run` `ID` is not present. We discussed that it should just write a warning since this causes issues for further steps in the workflow, but the wrapper itself can work. Any ideas on how this should then behave?

xtrojak: Covered in [273fe3f](https://github.com/RECETOX/galaxytools/pull/317/commits/273fe3fa52c256e5421d9ed263359018df6c50ee).

tools/recetox_aplcms/recetox_aplcms_merge_known_table.xml
```diff
@@ -0,0 +1,117 @@
...
+         ]]></configfile>
+    </configfiles>
+
+    <inputs>
+        <param name="metadata_file" type="data" format="parquet" label="Metadata table"
+               help="TBA" />
+        <param name="rt_file" type="data" format="parquet" label="RT table"
+               help="TBA" />
+        <param name="intensity_file" type="data" format="parquet" label="Intensity table"
+               help="TBA" />
```
xtrojak, line 53: @hechth any thoughts on how to briefly describe these?

hechth: Handled in https://github.com/RECETOX/galaxytools/pull/317/commits/01d9298d8675534ac076aa5ded1fc3abf30c8064

xtrojak: Actually fixed in [2c73963](https://github.com/RECETOX/galaxytools/pull/317/commits/2c739637658c753740ccddf72b1e8b81bb23bf7e), mentioned commit covered it for some other files.

PR #324, opened 2023-01-31
Msfinder version update

### LINKED ISSUE #323: Update recetox-msfinder to v3.5.2-rcx0
Update recetox-msfinder to latest version using the image on dockerhub.

The parameter interface can be simplified and the element checks removed, since we are not running any formula prediction, only peak annotation.

I think otherwise the parameters can stay the same so far. Or do you know that the MS1 tolerance is also used for the formula annotation?

zargham-ahmad:

- updated tool version
- updated docker image version
- removed/commented element check parameter

Closes #323

### DIFF
--- a/tools/recetox_msfinder/recetox_msfinder.xml
+++ b/tools/recetox_msfinder/recetox_msfinder.xml
@@ -1,4 +1,4 @@
-<tool id="recetox_msfinder" name="RECETOX MsFinder" version="@TOOL_VERSION@+galaxy1">
+<tool id="recetox_msfinder" name="RECETOX MsFinder" version="@TOOL_VERSION@+galaxy0">
     <description>Annotation of fragment peaks in mass spectral libraries.</description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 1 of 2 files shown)


tools/recetox_msfinder/recetox_msfinder.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="recetox_msfinder" name="RECETOX MsFinder" version="@TOOL_VERSION@+galaxy0">
+<tool id="recetox_msfinder" name="RECETOX MsFinder" version="@TOOL_VERSION@+galaxy1">
```
bgruening, line 1: this is not needed if you bump the TOOL_VERSION

zargham-ahmad: updated in [ae59da5](https://github.com/RECETOX/galaxytools/pull/324/commits/ae59da56e3ad0c0277d5dd9ae60d1d2aef73fd94)

hechth (approved):

PR #344, opened 2023-04-25
Create add_issue_to_project.yml

hechth:
Added template workflow which can be called from other repositories to add issues to the main project automatically


xtrojak:
@hechth Can you briefly describe how exactly we will use it?

hechth:
It is just a github action which can be used by other tools to automatically add issues to the main project

xtrojak (approved):
Sounds cool, looking forward to seeing it in practice!

PR #345, opened 2023-05-02
Updated recetox_aplcms_align_features with cluster

### LINKED ISSUE #340: recetox-aplcms: align features - pass number of allocated cores to the tool
Since this tool actually runs in parallel, we should pass the number of allocated cored for the galaxy job to the function - and also define that it should use for example 8 cores?

zargham-ahmad:

- Updated recetox_aplcms_align_features tool with cluster #340 

closes #340


xtrojak (approved):
I suppose this should work just fine. It is used the same way in the [testing tool](https://gitlab.ics.muni.cz/recetox/mass-spectrometry/umsa-galaxy/-/blob/main/files/galaxy/tools/testing.xml) installed on UMSA and it seems to be working.

BTW, in the future we could make resources configurable using [TPV](https://training.galaxyproject.org/training-material/topics/admin/tutorials/job-destinations/tutorial.html) (so you can actually select how many CPUs or memory you want to use for a particular job).

PR #346, opened 2023-05-04
Updated Labels in recetox_aplcms tools

### LINKED ISSUE #342: recetox-aplcms: change output object names to contain initial input name
Currently the objects are called `recetox-aplcms - generate features on data 1` while actually giving the name of data one here (and maybe the number of the dataset in brackets behind it) would be way more convenient.

zargham-ahmad:

- Updated Labels in recetox_aplcms tools #342 
- bump version

closes #342


tools/recetox_aplcms/recetox_aplcms_correct_time.xml
```diff
@@ -44,7 +44,7 @@
     </inputs>
 
     <outputs>
-        <data label="${tool.name} on ${on_string}" name="output_file" format="parquet" />
+        <data label="${($features_table.element_identifier).rsplit('.',1)[0]}.parquet" name="output_file" format="parquet" />
```
xtrojak, line 47: @zargham-ahmad Can you briefly explain what the label looked like before and after these changes? It is not clear from the code since some "Python magic" is done here. You can do it on an example.

zargham-ahmad: The element_identifier returns, for example, this "RCX_06_shortened.mzML", what I basically did is changed the ".mzML" with ".parquet" so the Name in the Dataset Information section will look like "RCX_06_shortened.parquet" for the recetox_aplcms tools we updated the labels for

xtrojak (approved):

tools/recetox_aplcms/recetox_aplcms_remove_noise.xml
```diff
@@ -39,7 +39,7 @@
     </inputs>
 
     <outputs>
-        <data label="${tool.name} on ${on_string}" name="output_file" format="parquet" />
+        <data label="${($input_file.element_identifier).rsplit('.',1)[0]}.parquet" name="output_file" format="parquet" />
```
bgruening, line 42: Could you maybe create an issue or some text with the underlying problem that y'all want to solve with this?

hechth (approved):
@bgruening I don't know what you mean with consistent collections so I'd just go ahead and merge this otherwise

bgruening:
Please go ahead. We can discuss this when we see each other. Thanks!

PR #362, opened 2023-05-24
Fixed bug in recetox-aplcms remove_noise

hechth:
There was a small bug which kept the CI from passing and therefor the tools were not pushed to the toolshed from the seems of it.


xtrojak:
Can you point to a particular CI run where this occurred? I don't think this should be an issue.

PR #364, opened 2023-05-31
RAMClustR: update to 1.3.0

### LINKED ISSUE #335: RAMClustR: update to 1.3.0
As soon as the RAMClustR changes are merged (https://github.com/cbroeckl/RAMClustR/pull/39) we can update the Galaxy tools to work with the newer version. This means we can implement another way to read data, which is passing a "data.frame" object. We can either accept anything that can be read into a dataframe under this (any tabular format or R objects) and see if it fulfills the requirements or we can explicitly accept only tabular files, i.e. parquet which is outputted from recetox-aplcms.

zargham-ahmad:

- updated the ramclustr package to 1.3.0
- added input option for recetox-aplcms
- rewrote the wrapper and use the individual steps
- added test for the df version

_note: normalization method qc and quantile will not work until the new version of ramclustr https://github.com/cbroeckl/RAMClustR/pull/44_

closes #335


hechth:
@zargham-ahmad please also add the option for qc normalization to the wrapper

hechth (approved):

PR #375, opened 2023-06-16
Add EDAM identifiers to recetox-aplcms suite

### LINKED ISSUE #368: Add EDAM identifiers to recetox-aplcms suite
recetox-aplcms

Input Data

    mass spectrometry data (2536)

Format

    mzML (3244)

Operation

    filtering (3937) (noise removal)
    feature extraction (generate features, recovery)
    peak detection (3215) (generate features, recovery)
    chromatographic alignment (3628) (rt correction step)
    alignment (2928) (alignment step)
    Qauntification (3799) (generate features, recovery)

Output

    feature table (1270)
    alignment (1916)
    expression data (2603)

hechth:
Fixes #368

### DIFF
--- a/tools/recetox_aplcms/macros.xml
+++ b/tools/recetox_aplcms/macros.xml
@@ -7,6 +7,17 @@
         </requirements>
     </xml>
 
+    <xml name="annotation">
+       <edam_topics>
+              <edam_topic>topic_3172</edam_topic>
+              <edam_topic>topic_0091</edam_topic>
+              <edam_topic>topic_3520</edam_topic>       
+       </edam_topics>
+       <xrefs>
+              <xref type="bio.tools">recetox-aplcms</xref>
+       </xrefs>
+    </xml>
+
     <xml name="creator">
         <creator>
             <person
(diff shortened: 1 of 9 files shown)


tools/recetox_aplcms/macros.xml
```diff
@@ -7,6 +7,17 @@
...
     </xml>
 
+    <xml name="annotation">
+       <edam_topics>
+              <edam_topic>topic_3172</edam_topic>
+              <edam_topic>topic_0091</edam_topic>
+              <edam_topic>topic_3520</edam_topic>       
+       </edam_topics>
+       <xrefs>
+              <xref type="bio.tools">recetox-aplcms</xref>
```
bgruening, line 17: IFF you have a bio.tools ID and the EDAM operations are stored under the bio.tools ID.

Galaxy is smart enought to annotate your tools automatically with EDAM using the bio.tools ID - means you do NOT need to add edam_topics ONLY the bio.tools ID.

hechth: Oh nice - I didn't know that!

What would be the case if a tool has multiple steps and we want to be precise? Like some steps in recetox-aplcms are peak detection and some are filtering - can we be explicit about that?

hechth: Also would be great if the documentation on bio.tools and EDAM and the Galaxy integration could be updated, or at least I didn't find this information there: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#edam-topics-and-operations

bgruening: > Oh nice - I didn't know that!
> 
> What would be the case if a tool has multiple steps and we want to be precise? Like some steps in recetox-aplcms are peak detection and some are filtering - can we be explicit about that?

Ideally, you would capture that in bio.tools. So I guess extra entries for every step? If this is unpractical, then what you have here works and will overwrite the bio.tools given operations.

bgruening: Sure, go for it. And maybe its different/easier for other tools.

zargham-ahmad (approved):

xtrojak (approved):

PR #376, opened 2023-06-16
Add EDAM identifiers to ramclustr

### LINKED ISSUE #369: Add EDAM identifiers to RAMClustR and RAMClustR define experiment
**RAMClustR**

Input Data

    mass spectrometry data (2536)
    Feature table (1270)
    Sample annotation (3113)

Format

    CSV (3752) 

Operation

    Imputation (3557)
    Standardisation and normalisation (3435)
    Clustering (3432)
    Correlation (3465)    

Output

    Mass spectrum (0943)
    Expression data (2603)

**RAMClustR define experiment**

Input Data

    Data (0006)

Format

    CSV (3752) 

Operation

    Generation (3429)  

Output

    Protocol (2531)

zargham-ahmad:
fixes #369

### DIFF
--- a/tools/ramclustr/ramclustr_define_experiment.xml
+++ b/tools/ramclustr/ramclustr_define_experiment.xml
@@ -1,4 +1,4 @@
-<tool id="ramclustr_define_experiment" name="RAMClustR define experiment" version="1.0.2" profile="21.09">
+<tool id="ramclustr_define_experiment" name="RAMClustR define experiment" version="1.0.2+galaxy0" profile="21.09">
     <description>Definition of experimental design used for record keeping and writing spectra data.</description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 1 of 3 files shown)


hechth (approved):

tools/ramclustr/ramclustr_define_experiment.xml
```diff
@@ -1,8 +1,12 @@
-<tool id="ramclustr_define_experiment" name="RAMClustR define experiment" version="1.0.2" profile="21.09">
+<tool id="ramclustr_define_experiment" name="RAMClustR define experiment" version="1.0.2+galaxy0" profile="21.09">
```
xtrojak, line 1: I think there is no reason to introduce this extra versioning layer since the tool itself exists only here. I would just bump the version to `1.0.3` or so.

hechth: I think it makes sense to separate the version of the python code from the Galaxy wrapper (for what it is worth?)

PR #378, opened 2023-06-22
Updated waveica for new outputs from recetox-aplcms

### LINKED ISSUE #365: check and update waveica for new outputs from recetox-aplcms and RAMClustR

zargham-ahmad:

- Added data transpose option so the new outputs from recetox-aplcms can be used
- Added verify_column_types function to check if the columns are of the correct type
- Added test and testdata for transpose case
- Fixed some param calls in the test (filter_length -> wavelet_length)
- added group var in waveica_singlebatch function which was not there but used
- Bump version

closes #365


hechth (commented):
LGTM - only add WaveICA to bio.tools and then add the crossref here and we're done :)

zargham-ahmad:
> LGTM - only add WaveICA to bio.tools and then add the crossref here and we're done :)

added in [a54098b](https://github.com/RECETOX/galaxytools/pull/378/commits/a54098b338e6df900ec886f7662ba47938a6d78c)

tools/waveica/waveica.xml
```diff
@@ -1,8 +1,9 @@
-<tool id="waveica" name="WaveICA" version="@TOOL_VERSION@+galaxy2" python_template_version="3.5">
+<tool id="waveica" name="WaveICA" version="@TOOL_VERSION@+galaxy3" python_template_version="3.5">
     <description>removal of batch effects for untargeted metabolomics data</description>
     <macros>
         <import>macros.xml</import>
     </macros>
+    <expand macro='annotation'/>
```
xtrojak, line 6: Keep consistent quotation marks type.

```suggestion
    <expand macro="annotation"/>
```

tools/waveica/waveica.xml
```diff
@@ -61,10 +64,12 @@
...
             </when>
             <when value="2">
+                <expand macro="input_data"/>
                 <section name="input_metadata" title="Input metadata table" expanded="true">
                     <param name="metadata" label="Input metadata" type="data" format="csv,tsv,parquet" help="" />
-                </section> 
-                <expand macro="input_data"/>
+                </section>
+                <param name = "transpose_feature_table" label="Transpose feature table" type="boolean" checked="false" 
+                truevalue="TRUE" falsevalue="FALSE" help="Transpose if feature names are -> column 1 and sampleNames are column" />
```
xtrojak, line 72: Somehow the `help` doesn't help.

xtrojak (approved):
Except minor issues looks good to me.

tools/waveica/waveica.xml
hechth: True. This could be improved a bit :D @zargham-ahmad

zargham-ahmad: Updated in [288aaae](https://github.com/RECETOX/galaxytools/pull/378/commits/288aaae211afc339670bc205845a4393665a8455)

zargham-ahmad: Is it better now? or maybe it seems more complicated now :D

xtrojak: Yeah, now it's actually a sentence :D but still takes a minute to understand. I'm thinking of it's even needed to explain, I suppose "transpose" is pretty clear. Anyway, maybe a more straightforward formulation: "Swap sample names with feature names as column headers (to fit aplcms outputs)."

zargham-ahmad: Yes, this looks good as well

PR #404, opened 2023-09-26
Added matchms add key tool

### LINKED ISSUE #248: Set metadata key in MSP to static value (change/edit)
The tool should take an input MSP file and take as parameters the name of the attribute and its value which will be assigned to all spectra records in the MSP.

Let's create a galaxy toool that takes an msp and 2 text inputs, one for the key and one for the value and which adds the key and value pair to every spectrum.

zargham-ahmad:

- Added matchms add key tool.
- Added test and testdata

closes #248


tools/matchms/matchms_add_key.xml
```diff
@@ -0,0 +1,70 @@
+<tool id="matchms_add_key" name="matchms add key" version="@TOOL_VERSION@+galaxy0" profile="21.05">
```
hechth, line 1:
```suggestion
<tool id="matchms_add_key" name="matchms add key" version="@TOOL_VERSION@+galaxy0" profile="21.09">
```

hechth (approved):

PR #437, opened 2023-11-20
Added tool: matchms_subsetting_tool

### LINKED ISSUE #431: matchms spectral library subsetting tool
We should implement a tool which allows extracting spectra from a library given their index or some unique metadata identifier like InChI or SPLASH or something comparable.

The tool should be made to work in 2 directions, (1) removing the spectra matching identifiers given in the file and (2) the opposite direction, only keeping the spectra matching the identifiers in the file.

Input for the Galaxy tool should be the (1) msp file to be filtered, (2) the mode as a select for `include` and `exclude` and the (3) list of identifiers in `tsv` or `csv` format to use for filtering. The header row in this list of identifiers states the key which will be used to match the identifiers in the list and the spectra.

Future extensions:

- [ ] Include option to filter based on index
- [ ] Automaticalle deduce the key based on the file format, e.g. for SMI files deduce that it is SMILES

zargham-ahmad:

- Implemented tool which allows extracting spectra from a library based on the provided list of identifier #431 
- added testdata

TODO
- [x] pending https://github.com/RECETOX/galaxytools/pull/442 merge

closes #431


tools/matchms/matchms_subsetting_tool.xml
```diff
@@ -0,0 +1,100 @@
...
+        </data>
+    </outputs>
+
+    <tests>
+        <test>
+            <param name="spectral_library" value="out_matchms_add_key.msp" ftype="msp"/>
+            <param name="mode" value="include"/>
+            <param name="list_of_identifiers" value="subsetting/identifier.csv" ftype="csv"/>
+            <output name="output" file="subsetting/subsetting_output.msp" ftype="msp"/>
+        </test>
```
hechth, line 88: Do we have a small msp file with only few spectra in the testdata? Then we could also test the exclude mode.

hechth (commented):
LGTM, only some small changes regarding the test and naming of the tool

tools/matchms/matchms_subsetting_tool.xml
zargham-ahmad: done in [114e150](https://github.com/RECETOX/galaxytools/pull/437/commits/114e15055931155fe37c6c10be558cd1a703dd8c)

hechth (approved):

PR #438, opened 2023-11-21
fix matchms outputs to existing types

bernt-matthias:
Otherwise, the data can not be used in downstream tools/visualizations (or the user needs to set the datatype).


hechth:
Thanks a lot @bernt-matthias !

PR #443, opened 2023-11-23
Matchms networking: error message

bernt-matthias:
The new test currently gives a python Traceback that is hard to read by users. Can this be improved in the wrapper / matchms?

### DIFF
--- a/tools/matchms/matchms_networking.xml
+++ b/tools/matchms/matchms_networking.xml
@@ -17,31 +17,25 @@
     </requirements>
 
     <command detect_errors="aggressive"><![CDATA[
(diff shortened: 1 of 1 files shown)


bgruening (approved):

bernt-matthias:
Please don't merge yet. Would like to see first if we can get a better error message in stderr? Ping @hechth

hechth:
@bernt-matthias soo whats up with this here? Anything I can do?

bernt-matthias:
Question is if there should/can be a proper error message instead of the `ValueError`.

hechth (approved):

PR #451, opened 2023-11-29
Updated recetox-msfinder output and version

zargham-ahmad:

- updated msfinder docker tag
- updated tool output to either be an actual annotation or log file
- added test and testdata

### DIFF
--- a/tools/recetox_msfinder/macros.xml
+++ b/tools/recetox_msfinder/macros.xml
@@ -2,7 +2,7 @@
     <token name="@TOOL_VERSION@">v3.5.2</token>
     <xml name="requirements">
         <requirements>
-            <container type="docker">recetox/recetox-msfinder:v3.5.2-rcx0</container>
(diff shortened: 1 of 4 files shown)


tools/recetox_msfinder/macros.xml
```diff
@@ -2,7 +2,7 @@
     <token name="@TOOL_VERSION@">v3.5.2</token>
     <xml name="requirements">
         <requirements>
-            <container type="docker">recetox/recetox-msfinder:v3.5.2-rcx0</container>
+            <container type="docker">recetox/recetox-msfinder:v3.5.2-rcx1</container>
```
hechth, line 5: Putting the docker image tag -rcx1 into the tool version was not legal if I remember it correctly?

hechth (approved):

tools/recetox_msfinder/macros.xml
zargham-ahmad: yes

PR #468, opened 2023-12-04
Adding the ipaPy2 tool suite

hechth:
The first stub which doesn't take any parameters at this point is already available and can be used for further implementations. We still need to pass additional parameters to the tool.


hechth:
@francescodc87 we still need to fill in the `owner` field in the shed.yml. This can either be your university or we can put in recetox. This means under who's account the tool will be stored in the toolshed, which is used for versioning and also deprecating tools. We currently have all our tools under one organizational account - depending on whether you intend to add more tools it would make sense to add your uni as an owner or us.

francescodc87:
> @francescodc87 we still need to fill in the `owner` field in the shed.yml. This can either be your university or we can put in recetox. This means under who's account the tool will be stored in the toolshed, which is used for versioning and also deprecating tools. We currently have all our tools under one organizational account - depending on whether you intend to add more tools it would make sense to add your uni as an owner or us.

@hechth can there be more than one owners? It would make sense to have both University of Liverpool and University of Manchester as owners

bgruening:
@francescodc87 this TS owner is just the TS account and should not be used for attribution.

Please use https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-creator-organization to credit your universities.

tools/ipapy2/ipapy2_clustering.xml
```diff
@@ -0,0 +1,60 @@
...
+        <import>macros.xml</import>
+    </macros>
+    
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">ipapy2</requirement>
+        <requirement type="package" version="8.0.1">click</requirement>
+    </requirements>
+
+    <command detect_errors="exit_code"><![CDATA[
+        python3 ${__tool_directory__}/ipapy2_clustering.py --i '${intensity_table}' --o '${output}'
```
bgruening, line 12:
```suggestion
        python3 '${__tool_directory__}/ipapy2_clustering.py' --i '${intensity_table}' --o '${output}'
```

tools/ipapy2/ipapy2_clustering.xml
```diff
@@ -0,0 +1,60 @@
+<tool id="ipapy2_clustering" name="ipaPy2 clustering" version="@TOOL_VERSION@+galaxy0" python_template_version="3.5" profile="21.05">
```
bgruening, line 1:
```suggestion
<tool id="ipapy2_clustering" name="ipaPy2 clustering" version="@TOOL_VERSION@+galaxy0" profile="21.05">
```

Description is missing atm

tools/ipapy2/ipapy2_clustering.py
```diff
@@ -0,0 +1,16 @@
+import click
+import pandas as pd
+from ipaPy2 import ipa
```
bgruening, line 3: if we put this small function into a `<configfile>` or use argparse we could save the click dependency and the additional container.

tools/ipapy2/ipapy2_clustering.xml
```diff
@@ -0,0 +1,60 @@
...
+            <param name="intensity_table" value="minimal_input.csv"/>
+            <output name="output" file="clustering.csv"/>
+        </test>
+    </tests>
+
+    <help><![CDATA[
+    Before using the ipaPy2 package, the processed data coming from an untargeted metabolomics experiment must be properly prepared.
+    The data must be organized in a pandas dataframe containing the following columns:
+
+    - **ids**: an unique numeric id for each mass spectrometry feature feature
```
francescodc87, line 34:
```suggestion
    - **ids**: an unique numeric id for each mass spectrometry feature
```

hechth:
@francescodc87 the main question is whether you want to maintain the wrappers yourself or not - I guess depending on that we should choose the toolshed owner. As Bjoern said, this is not about representation or credit - this is handled differently using the tags in the XML file.

francescodc87:
@hechth I see. Having recetox as owner is fine by me then

PR #475, opened 2023-12-11
matchms_filtering: Added derive_precursor_mz_from_parent_mass filter

### LINKED ISSUE #457: Add `derive_precursor_mz_from_parent_mass` to matchms filtering

zargham-ahmad:

- Added derive_precursor_mz_from_parent_mass filter
- Added test and testdata 
- bump galaxy version

_Note: I had to convert parent_mass explicitly to float before applying the filter otherwise, matchms was giving the error._

closes #457


hechth:
@zargham-ahmad the conversion to `int` can't be there as the mass can also be a non integer number. Does it work when converting to float? Otherwise we need to see in more detail what the problem is.

tools/matchms/matchms_filtering_wrapper.py
```diff
@@ -84,6 +89,11 @@ def main(argv):
         if args.reduce_to_top_n_peaks:
             spectrum = reduce_to_number_of_peaks(spectrum_in=spectrum, n_max=args.n_max)
 
+        if args.derive_precursor_mz_from_parent_mass:
+            spectrum.set("parent_mass", int(float(spectrum.get('parent_mass'))))
```
hechth, line 93: This line has to be removed to keep the feature functional.

hechth (changes requested):

zargham-ahmad:
> @zargham-ahmad the conversion to `int` can't be there as the mass can also be a non integer number. Does it work when converting to float? Otherwise we need to see in more detail what the problem is.

yes, it seems to work with converting to float

tools/matchms/matchms_filtering_wrapper.py
zargham-ahmad: removed conversion to int in 9e579c4

tools/matchms/matchms_filtering.xml
```diff
@@ -55,6 +55,10 @@
             #if $require_inchi_is_true == "TRUE"
                 -require_inchi \
             #end if
+            #if $derive_precursor_mz_from_parent_mass.is_true == "TRUE"
+                -derive_precursor_mz_from_parent_mass \
+                --estimate_from_adduct "$derive_precursor_mz_from_parent_mass.estimate_from_adduct" \
+            #end if
```
hechth, line 61:
```suggestion
            #if ${derive_precursor_mz_from_parent_mass.is_true} == "TRUE"
                -derive_precursor_mz_from_parent_mass \
                --estimate_from_adduct "${derive_precursor_mz_from_parent_mass.estimate_from_adduct}" \
            #end if
```

hechth (approved):
looks good

hechth (approved):

PR #504, opened 2024-03-19
Change output csv types to tsv 

### LINKED ISSUE #439: Change all output table types to tabular unless they are parquet
- [x] matchms metadata export
- [x] RAMClustR abundance table
- [x] MSMetaEnhancer on tabular data
- [x] RIAssigner on tabular data
- [x] matchms scores formatter
- [x] biotransformer
- [x] rem_complexes

zargham-ahmad:
updated csv output types for the below tools to tsv and updated test.

-  matchms metadata export
-  RAMClustR abundance table
-  MSMetaEnhancer on tabular data
-  RIAssigner on tabular data
-  matchms scores formatter
-  biotransformer
-  rem_complexes

closes #439


tools/msmetaenhancer/msmetaenhancer.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="msmetaenhancer" name="MSMetaEnhancer" version="@TOOL_VERSION@+galaxy3">
+<tool id="msmetaenhancer" name="MSMetaEnhancer" version="@TOOL_VERSION@+galaxy4">
```
wverastegui, line 1:
```suggestion
<tool id="msmetaenhancer" name="MSMetaEnhancer" version="@TOOL_VERSION@+galaxy4" profile="21.09">
```

wverastegui (changes requested):
Just a minor comment, otherwise it looks good to me.

tools/ramclustr/ramclustr_wrapper.R
```diff
@@ -3,9 +3,9 @@ store_output <- function(ramclustr_obj,
                          output_spec_abundance,
                          msp_file) {
   RAMClustR::write.msp(ramclustr_obj, one.file = output_merge_msp)
-  write.csv(ramclustr_obj$SpecAbund,
+  write.table(ramclustr_obj$SpecAbund,
     file = output_spec_abundance,
-    row.names = TRUE, quote = FALSE
+    row.names = TRUE, quote = FALSE, col.names = NA, sep = "\t"
```
hechth, line 8: Doesn't `col.names = NA` lead to the column names not being written to the file?

tools/rem_complex/rem_complex.xml
```diff
@@ -23,7 +23,12 @@
         <param name="input" format="inchi,smi,csv" type="data" help="Accepted input formats: CSV, SMI, and InChI."/>
```
hechth, line 23: Would be good to also support tsv as input

tools/riassigner/riassigner_from_comment.xml
```diff
@@ -15,7 +15,11 @@
 
     <command detect_errors="exit_code"><![CDATA[
```
hechth, line 16: I have the feeling that the result writing could be simplified here, to avoid having to use `mv` etc.

tools/riassigner/riassigner_from_comment.xml
```diff
@@ -46,13 +54,19 @@ query.write('results.${input.ext}')
...
 
     <outputs>
-        <data format_source="input" name="output" label="${on_string} with ${format} RI"/>
+        <data format="tsv" name="output" label="${on_string} with ${format} RI">
+            <change_format>
+                <when input="input.ext" value="msp" format="msp" />
+                <when input="input.ext" value="mgf" format="mgf" />
+                <when input="input.ext" value="parquet" format="parquet" />
+            </change_format>
+        </data>
```
hechth, line 63: nice!

hechth (commented):
So far so good, just some minor comments which would make the things still a bit nicer.

tools/ramclustr/ramclustr_wrapper.R
zargham-ahmad: in write.table function it does not have this meaning as per the [documentation](https://www.rdocumentation.org/packages/utils/versions/3.6.2/topics/write.table)
_"By default there is no column name for a column of row names. If col.names = NA and row.names = TRUE a blank column name is added, which is the convention used for CSV files to be read by spreadsheets."_

tools/riassigner/riassigner_from_comment.xml
zargham-ahmad: I haven't really changed the command much as It is a bit tricky because the writer expects the filename with an extension, so I can't pass the tool output directly to it. Also cannot use the from_work_dir because we have multiple extensions.

tools/rem_complex/rem_complex.xml
zargham-ahmad: done in [530212f](https://github.com/RECETOX/galaxytools/pull/504/commits/530212f6efa29c3aa625d93e557b2ae0535e988b)

zargham-ahmad:
@hechth CI was failing because of the incompatibility between the upload and download action version _"Artifacts created with versions v3 and below are not compatible with the v4 actions. Uploads and downloads must use the same major action versions."_ [link](https://github.com/actions/upload-artifact/releases/tag/v4.0.0). I updated the download-artifact to v4, which resolved the issue

hechth (approved):

PR #517, opened 2024-04-17
Implemented dedicated require filters matchms tool

### LINKED ISSUE #507: Implement explicit `matchms remove spectra` tool
Since the matchms filtering tool got quite cluttered it would be nice to move the `require` filters to a dedicated tool.

This tool should have one multiselect as main input where the individual metadata fields are chosen and where spectra which don't have this metadata information are removed.

The options that should be supported are
* SMILES
* InChI
* InChIKey
* Formula
* Retention time
* Retention index
* Precursor mz
* Valid Annotation
* Compound Name

A second input could be a boolean that can output optionally the spectra that have been removed. Another option would be to output a logfile using a identifier (`compound_name` preferably) or an index if this is not present and to explain because of which missing metadata entry this was removed.

If some of the filters are not available in matchms, they should be added.

The `Valid Annotation` option should get a help text explaining the functionality as it is implemented in matchms.

zargham-ahmad:

- Implemented dedicated require filters matchms tool
- updated matchms_filtering
- added test and testdata

closes #507


hechth:
@zargham-ahmad latest matchms version 0.25.0 is on bioconda :)

PR #521, opened 2024-04-22
Isolib tool first version

### LINKED ISSUE #427: Implement tool to generate an isotopic pattern library
To actually match detection of isotopic peaks in the data, a tool to generate an isotopic pattern library from a compound library should be implemented. The tool should consume a table with at minimum `formula` and `rt`, `ionmode` and `adduct` and create an MSP file with all other columns in the metadata and the isotopic pattern for this formula created and converted into peaks using the adduct and ionmode information.

hechth:
This PR adds a new tool to calculate isotopic pattern and adduct libraries given a list of compounds.
It is based on enviPat and the RforMassSpectrometry toolbox.

Required columns for the compound table are only `formula` and `name`. `rt` is used if present and added to the spectra to integrate with retention time filtering with matchms.

Future work

- [ ] isolib: add additional adducts
- [ ] isolib: allow merging spectra for multiple adducts?

Closes #427


tools/isolib/isolib.xml
```diff
@@ -0,0 +1,72 @@
...
+        <edam_operation>operation_3632</edam_operation>
+    </edam_operations>
+    <requirements>
+        <requirement type="package" version="1.10.0">bioconductor-metabocoreutils</requirement>
+        <requirement type="package" version="1.12.0">bioconductor-spectra</requirement>
+        <requirement type="package" version="1.6.0">bioconductor-msbackendmsp</requirement>
+        <requirement type="package" version="2.6">r-envipat</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        Rscript -e 'source("${__tool_directory__}/isolib.R")' '${input_file}' '${adducts}' '${threshold}' '${append_adduct}' '${isotope_library}'
```
bgruening, line 24: I don't understand why source() is needed, I think you can remove that.

hechth: true

PR #522, opened 2024-04-23
Update isolib.xml

hechth:
making compatible with profile 21.09

### DIFF
--- a/tools/isolib/isolib.xml
+++ b/tools/isolib/isolib.xml
@@ -1,4 +1,4 @@
-<tool id="isolib" name="isolib" version="1.0.0+galaxy0" profile="21.09">
+<tool id="isolib" name="isolib" version="1.0.0+galaxy1" profile="21.09">
     <description>create an isotopic pattern library for given compounds and adducts</description>
     <creator>
         <person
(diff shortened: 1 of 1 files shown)


tools/isolib/isolib.xml
```diff
@@ -21,7 +21,7 @@
         <requirement type="package" version="2.6">r-envipat</requirement>
     </requirements>
     <command detect_errors="exit_code"><![CDATA[
-        Rscript ${__tool_directory__}/isolib.R '${input_file}' '${adducts}' '${threshold}' '${append_adduct}' '${isotope_library}'
+        Rscript ${__tool_directory__}/isolib.R '${input_file}' '${ionization.adducts}' '${threshold}' '${append_adduct}' '${isotope_library}'
```
bgruening, line 24: The tool_dir needs to be single quotes.

hechth: It also works like this - but ok will change the path to single quotes

bgruening: Put a space in path where you install your tools and it will stop working :-)

PR #525, opened 2024-04-24
isolib: fix rt column reading and writing

### LINKED ISSUE #524: isolib: fix rt column reading and writing

hechth:
Fixes #524


tools/isolib/isolib.R
```diff
@@ -48,7 +48,7 @@ main <- function() {
     )
 
     if ("rt" %in% colnames(compound_table)) {
-      spectra_df$retention_time <- compound_table$rt
+      spectra_df$retention_time <- as.numeric(sub(",", ".", compound_table$rt, fixed = TRUE))
```
bgruening, line 51: is this really how you do it in R? Can you not set a locale setting somewhere? So that R respects . vs ,?

hechth: We can add a dependency to readr for example to include a more sophisticated table reading function, sure

bgruening: I'm just surprised. Those things normally depend on your `locale` setting of your environment.

hechth: yeah - but how to tell it to accept both?

hechth: how does usegalaxy.eu and usegalaxy.org handle these interoperabilities? :D

bgruening: We do set one `locale` for all tools. US I think. So that all tools are compatible ...

PR #553, opened 2024-06-03
Updated matchms remove spectra tool

hechth:
* removing now spectra with 0 value as retention index
* Fixed bug related to variables and inchikey removal - the string matching logic had some bugs
* simplified logic


zargham-ahmad (approved):
Looks good to me!

Issue #534, opened 2024-05-07
Retain molecule name in optimized XYZ file 


wverastegui:
Description:
After running an optimization using the xTB molecular optimization tool, the original molecule name in the XYZ file is lost in the output file. It would be beneficial if the tool could retain the original molecule name in the optimized XYZ file by default, while optionally preserving the information added by xTB.

Expected Behavior:
The optimized XYZ file should keep the molecule name from the original XYZ file.

PR #557, opened 2024-06-11
Added boolean to retain molecule name 

wverastegui:
This PR adds a Boolean option to retain the original molecule name in the optimized XYZ file if set to "YES". By default, it is set to "NO", preserving the information added by xTB.

closes https://github.com/RECETOX/galaxytools/issues/534


tools/xtb/xtb_molecular_optimization.xml
```diff
@@ -24,29 +35,48 @@
...
+            </output>
+        </test>
+        <test>
+            <param name="coordinates" value="initial_mol_no_molname.xyz" ftype="xyz"/>
+            <param name="replace_second_line" value="TRUE" />
+            <output name="output">
                 <assert_contents>
-                    <has_text text="$coord"/>
-                    <has_n_lines min="26"/> 
+                    <has_text text=" "/>
```
hechth, line 67: This merely tests if it has a space. There is I think an assert to check a the content of a specific line - this could be useful here.

hechth: I think the previous check was maybe even more concise in this place.

hechth (commented):
The new tests look good, except for the one checking for an empty string - that one should be improved. What was wrong with the previous test for this case?

tools/xtb/xtb_molecular_optimization.xml
wverastegui: I think <not_has_text text="energy"/> fits to this case as well [b4eb0e6](https://github.com/RECETOX/galaxytools/pull/557/commits/b4eb0e6309b132c5da5bc80e11322cc01b12b454).

hechth (approved):

PR #594, opened 2024-10-16
signal-to-noise separated into ratio and noise in IsoFiltR

### LINKED ISSUE #588: MFAssignR: separate signal-to-noise input into ratio and noise in IsoFiltR
In the IsoFiltR function, separate the signal-to-noise input into `ratio` and `estimated_noise` from the KMDnoise function (similarly as it's done in other functions).

KristinaGomoryova:
Signal-to-ratio parameter is now separated into `kmdn` * `SN_ratio` similarly as it's done in other functions (e.g. MFAssign). 

closes #588

### DIFF
--- a/tools/mfassignr/mfassignr_isofiltr.xml
+++ b/tools/mfassignr/mfassignr_isofiltr.xml
@@ -1,4 +1,4 @@
-<tool id="mfassignr_isofiltr" name="MFAssignR IsoFiltR" version="@TOOL_VERSION@+galaxy0" profile="23.0">
+<tool id="mfassignr_isofiltr" name="MFAssignR IsoFiltR" version="@TOOL_VERSION@+galaxy1" profile="23.0">
     <description>Separates likely isotopic masses from monoisotopic masses in a mass list</description>
     <macros>
         <import>macros.xml</import>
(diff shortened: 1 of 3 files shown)


tools/mfassignr/mfassignr_isofiltr.xml
hechth, line 1: Please update the galaxy tool wrapper version. Otherwise looks good - does the test need updating?

KristinaGomoryova: resolved in 2778854dfdd6f599cf3f1fb82747aa443e69178b

KristinaGomoryova: I think in this case we don't need to update the test, there are only datasets which don't change in this case

hechth (approved):

hechth:
@KristinaGomoryova is this the only tool where this was still combined? Or aren't there more of them yet?

KristinaGomoryova:
Well, this is somewhat related https://github.com/RECETOX/MFAssignR/issues/63

hechth:
Please make the same changes also to the SNplot tool where we currently have the cutoff value as a single parameter.

KristinaGomoryova:
> Please make the same changes also to the SNplot tool where we currently have the cutoff value as a single parameter.

Resolved in [cc5048d](https://github.com/RECETOX/galaxytools/pull/594/commits/cc5048dec07c4ac8905b8d25bfeb0d30682e2cfd)

hechth (approved):

PR #596, opened 2024-10-21
Force input of rt correction in recetox-aplcms to be a collection

hechth:


tools/recetox_aplcms/recetox_aplcms_correct_time.xml
```diff
@@ -37,9 +37,9 @@
     </configfiles>
 
     <inputs>
-        <param label="Input clustered features table" name="features_table" type="data" format="parquet"
-               help="Mass spectrometry clustered features table." />
-        <param label="Input template features table" name="template" type="data" format="parquet"
+        <param label="Input clustered feature tables" name="features_table" type="data_collection" collection_type="list" format="parquet"
```
bgruening, line 40: why is multiple=true not enough?

hechth: It kind of needs to be a collection - otherwise It doesn't make much sense. Also, the way how I want to use the tool in a workflow, if my one input is a list of lists, I'd expect the other one to be a list - this is currently not possible in the workflow editor.

PR #610, opened 2024-11-14
Adding new generation of table manipulation tools

hechth:
This PR introduces a set of new tabular file handling tools which are a bit more modern than what is available in the core galaxy tools.

The tools are based on pandas and/or scipy and all have a python CLI using argparse and should support parquet, csv, tsv and tabular files.

- [x] arithmetic operations
- [x] interpolate
- [x] rename columns
- [x] transform on columns - log, ld, ln, sqrt, e, abs, floor, ceil - basically anything that is of the form f(x) = y
- [ ] compute masks
- [ ] scale column subset
- [ ] summary

@bgruening and @bernt-matthias and  @mvdbeek please give some feedback if you think this design works well or whether it should be done differently. I'm planning to add few more tools to this to cover many use-cases in various -omics disciplines.

Also @KristinaGomoryova please try the tools locally to see if this is roughly what you had in mind.

Also, please let me know whether you would see these tools rather in tools-iuc and if you have a better idea for naming this collection of tools, I'm open to suggestions.


bgruening:
Do you know this tool? https://github.com/galaxyproject/tools-iuc/tree/main/tools/table_compute

hechth:
Yes I do ... but the tool is too complex in the form for what it can do and not very flexible. You can't choose specific columns etc. and overall I think it doesn't really cover the use cases. We actually adopted this schema from KNIME which provides very nice tools and operations to perform over tables.

tools/tables/.shed.yml
```diff
@@ -0,0 +1,16 @@
+name: tables
+owner: recetox
+remote_repository_url: "https://github.com/RECETOX/galaxytools/tree/master/tools/analysis"
```
bernt-matthias, line 3: The URLs would need fixes

tools/tables/table_pandas_arithmetics.xml
```diff
@@ -0,0 +1,55 @@
+<tool id="pandas_arithmetics" name="pandas arithmetics" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="20.01" license="MIT">
```
bernt-matthias, line 1: For this tool I would think that an extension for table compute is more appropriate. The possibility to select specific columns / rows for arithmetic operations would definitely make sense.

tools/tables/table_scipy_interpolate.xml
```diff
@@ -0,0 +1,88 @@
...
+        --xnew_col $xnew_col
+        --method '$method'
+        --output_dataset '$output_dataset' '$output_dataset.ext'
+        ]]></command>
+
+    <inputs>
+        <param argument="--reference_dataset" type="data" format="tabular,csv,tsv,parquet" label="Reference data" help="Reference dataset to use fopr the interpolation" />
+        <param name="x_col" type="data_column" data_ref="reference_dataset" use_header_names="true" label="x (reference)" help="Column from the reference dataset to use as X axis for the interpolator."/>
+        <param name="y_col" type="data_column" data_ref="reference_dataset" use_header_names="true" label="y (reference)" help="Column from the reference dataset to use as Y axis for the interpolator."/>
+        <param argument="--query_dataset" type="data" format="tabular,csv,tsv,parquet" label="Query dataset" help="Query dataset for which to interpolate the values." />
```
bernt-matthias, line 38: Are there any constraints regarding the range of the x values of the reference and the query?

bernt-matthias (commented):
I share the concerns of @bgruening and think that (in particular in case of the arithmetics and rename tool) it might be better to extend/improve existing tools.

bernt-matthias:
It could be an option to create 1-tool workflows for the tasks where we all the complexity of the the underlying tool gets hidden.

hechth:
@bernt-matthias yes, that is possible - though workflows don't support the column select input type properly, so you can't dynamically select the columns, which massively reduces user friendliness.

bernt-matthias:
> @bernt-matthias yes, that is possible - though workflows don't support the column select input type properly, so you can't dynamically select the columns, which massively reduces user friendliness.

Got it. 

> We actually adopted this schema from KNIME which provides very nice tools and operations to perform over tables.

This is a point I totally support :)

hechth:
@bernt-matthias and @bgruening I would like to move this forward but I'm afraid there won't be a community consensus any time soon. I would recommend that we will pilot the tools locally, moving them to the toolshed but keeping them out of usegalaxy.* until we have some more experience with what is really needed and if the usability and everything is how we expected it.

hechth:
@bernt-matthias thanks a lot, that would be very helpful!

rmassei:
Hi @bernt-matthias and @hechth!
After reviewing the PR, I have a few comments:

- The behavior of [table_pandas_arithmetics.xml](https://github.com/RECETOX/galaxytools/pull/610/files#diff-bc204851521f1f2971ddbf2cdfd183809761000180243279e820687b137058a0) and [tools/tables/table_pandas_transform.xml](https://github.com/RECETOX/galaxytools/pull/610/files#diff-990dff04c6b5676845dce19b5a932579266e3c8fccae0709e702e0fb2a5fd407) can already be replicated using the existing [table compute tool](https://github.com/galaxyproject/tools-iuc/tree/main/tools/table_compute). While it would require additional steps (e.g., performing the operation on a single column via cut and replace), the output would be identical. Since this operation is likely to be frequently performed by users, it might make more sense to enhance the existing table compute tool to include this functionality, especially since it is also based on pandas.

- A similar comment applies to [tools/tables/table_pandas_rename_column.xml](https://github.com/RECETOX/galaxytools/pull/610/files#diff-568eb0a35dbfcf57bd40a20eaf4c99c588a80ed93f6f8210c92ff2d6f22b13d7). This functionality can already be achieved using a simple workflow or by extending the table compute tool (or even the add_column_headers tool) to include this feature.

By design, KNIME offers a broader selection of data wrangling tools, which might indeed feel more user-friendly. However, if a tool with similar functionality already exists in Galaxy (such as the table compute tool) and is likely already integrated into some workflows, wouldn't it be more practical to enhance the functionality of the existing tool and provide more user-friendly documentation in its help section?

tools/tables/table_pandas_arithmetics.xml
hechth: Having the multi select for the rows works, but it would probably have to provide both options

PR #624, opened 2024-12-04
Added expression tool to ease parameter sweeps of recetox-aplcms on Galaxy

hechth:
This PR adds an expression tool to the recetox_aplcms suite that parses the parameters sweeped during peak detection.
This tool can be extended to contain more parameters in the future.

### DIFF
--- a/tools/recetox_aplcms/recetox_aplcms_parse_parameters.xml
+++ b/tools/recetox_aplcms/recetox_aplcms_parse_parameters.xml
@@ -5,11 +5,16 @@
         <token name="@TOOL_VERSION@">0.1.0</token>
         <token name="@VERSION_SUFFIX@">0</token>
     </macros>
-    <!-- TODO: please annotate this tool with topics and operations from http://edamontology.org -->
-    <!-- TODO: for more information see: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#edam-topics-and-operations -->
 
-    <requirements>
-    </requirements>
+    <edam_operations>
+        <edam_operation>operation_1812</edam_operation>
+    </edam_operations>
+
+    <edam_topics>
+        <edam_topic>topic_3316</edam_topic>
+        <edam_topic>topic_3071</edam_topic>
+    </edam_topics>
+
     <expression type="ecma5.1"><![CDATA[{
         // Read the input CSV file
         var data = $job.row.contents;
--- a/tools/recetox_aplcms/recetox_aplcms_parse_parameters.xml
+++ b/tools/recetox_aplcms/recetox_aplcms_parse_parameters.xml
@@ -1,0 +1,147 @@
+<tool id="recetox_aplcms_parse_parameters" name="recetox-aplcms - parse parameters" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="20.01" license="MIT" tool_type="expression">
+    <description>tool to parse parameters from a csv row for parameter searches, to be used in workflows only.</description>
+
+    <macros>
+        <token name="@TOOL_VERSION@">0.1.0</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <!-- TODO: please annotate this tool with topics and operations from http://edamontology.org -->
+    <!-- TODO: for more information see: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#edam-topics-and-operations -->
+
+    <requirements>
+    </requirements>
+    <expression type="ecma5.1"><![CDATA[{
+        // Read the input CSV file
+        var data = $job.row.contents;
+
+        // Split the data into lines
+        var lines = data.trim().split('\n');
+
+        // Get the header and the first data row
+        var header = lines[0].split(',');
+        var row = lines[1].split(',');
+
+        // Create a dictionary with column names as keys and row values as values
+        var result = {};
+        for (var i = 0; i < header.length; i++) {
+            result[header[i]] = row[i];
+        }
+        
+        var weighting;
+        if (result['weighting'] == 'true') {
+            weighting = true;
+        } else {
+            weighting = false;
+        }
+
+        return {
+            'min_run': parseFloat(result['min_run']),
+            'sigma_lower': parseFloat(result['sigma_lower']),
+            'sigma_higher': parseFloat(result['sigma_higher']),
+            'min_sd': parseFloat(result['min_sd']),
+            'max_sd': parseFloat(result['max_sd']),
+            'mz_tol': parseFloat(result['mz_tol']),
+            'weighting': weighting,
+            'min_pres': parseFloat(result['min_pres']),
+            'group_threshold': parseFloat(result['group_threshold'])
+        };
+    }]]></expression>
+
+    <inputs>
+        <param argument="--row" type="data" format="csv" label="Row of a CSV with the parameters, with header." help="Header has to be: [min_run,sigma_lower,sigma_higher,min_sd,max_sd,mz_tol,weighting,min_pres,group_threshold
+]" load_contents="64000"/>
+    </inputs>
+    <outputs>
+        <output name="min_run" type="float" label="min_run of ${on_string}" from="min_run" />
+        <output name="sigma_lower" type="float" label="sigma_lower of ${on_string}" from="sigma_lower" />
+        <output name="sigma_higher" type="float" label="sigma_higher of ${on_string}" from="sigma_higher" />
+        <output name="min_sd" type="float" label="min_sd of ${on_string}" from="min_sd" />
+        <output name="max_sd" type="float" label="max_sd of ${on_string}" from="max_sd" />
+        <output name="mz_tol" type="float" label="mz_tol of ${on_string}" from="mz_tol" />
+        <output name="weighting" type="boolean" label="weighting of ${on_string}" from="weighting" />
+        <output name="min_pres" type="float" label="min_pres of ${on_string}" from="min_pres" />
+        <output name="group_threshold" type="float" label="group_threshold of ${on_string}" from="group_threshold" />
+    </outputs>
+    <tests>
+        <!-- Hint: You can use [ctrl+alt+t] after defining the inputs/outputs to auto-scaffold some basic test cases. -->
+         
+    <test>
+        <!--TODO: auto-generated test case. Please fill in the required values-->
+        <param name="row" value="parse_parameters/test.csv"/>
+        <output name="min_run">
+            <assert_contents>
+                <has_text text="3.1"/>
+            </assert_contents>
+        </output>
+        <output name="sigma_lower">
+            <assert_contents>
+                <has_text text="0.5"/>
+            </assert_contents>
+        </output>
+        <output name="sigma_higher">
+            <assert_contents>
+                <has_text text="1.5"/>
+            </assert_contents>
+        </output>
+        <output name="min_sd">
+            <assert_contents>
+                <has_text text="0.1"/>
+            </assert_contents>
+        </output>
+        <output name="max_sd">
+            <assert_contents>
+                <has_text text="0.9"/>
+            </assert_contents>
+        </output>
+        <output name="mz_tol">
+            <assert_contents>
+                <has_text text="0.01"/>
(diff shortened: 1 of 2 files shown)


hechth:
maybe @mvdbeek has an opinion on this as well? I didn't figure how to read the passed csv file in the ECMA script but I tried to include some validation logic on the file contents.

hechth:
> > how to read the passed csv file in the ECMA script b
> 
> that's the point of it, you can't do I/O while executing ecmascript making this a safe choice for evaluating expressions. `load_contents="64000"` means your loading up max 64000 characters from your file.

Thanks, ok that makes sense. So do you think that the tool is fine like this? Probably the number of characters can be significantly reduced but apart from that? @bgruening has some concerns about this

mvdbeek:
Honestly this is brilliant and what I hope will happen more in the future. What are the concerns ?

mvdbeek:
When I say more I hope we can just let people write these expressions directly without having to write a separate standalone tool. The UI will be tricky to get right, but if you only do the one exact thing you need to do I think this reduces workflow and tool complexity a lot.

bgruening:
My concern was that it is too specific to the aplcms use-case. I would write this tool more generally and get it upstream or in IUC. I will look at it more tomorrow.

hechth:
### Review

**How this was validated** — so you can weigh the claims: I ran `planemo lint` and `planemo test` against a real Galaxy (25.1). The happy-path test **passes** (`exit_code: 0`, `All 1 test(s) executed passed`). I also ran the expression through Galaxy's own evaluation path (`cwlNodeEngine.js` wrapping, with a `$job` payload reconstructed from `wrapped_json.py`) to probe edge cases, and checked the output/metadata plumbing in Galaxy source. Everything below marked with a source is from an executed run or from reading Galaxy/IUC source, not from reading the XML alone.

### The non-obvious parts are correct

Worth saying explicitly, because these are the parts that are easy to get wrong:

- **The `$`-references resolve.** `ExpressionTool` binds `var $job = <job payload>` (`js_engine.py::build_evaluate_program`; `cwlNodeEngine.js:35` in 25.1). `_hda_to_object` emits `{file_ext, file_size, name, metadata, src}` plus `contents` when `load_contents` is set (`wrapped_json.py:131-139`), and `BaseCSV.set_meta` sets `column_names = header_row` as a genuine list of strings for CSV HDAs (`tabular.py:1482`). HDAs keep lists as lists — only the *library*-dataset `to_dict` variants join lists with `", "`.
- **The leading `{` in the CDATA is load-bearing and correct.** Galaxy treats the script as a function body *only* if `script[0] == "{"`; otherwise it wraps it as a bare expression and the top-level `return` is a syntax error. Nobody reading this XML would know that — it deserves a comment.
- `load_contents="64000"` is the canonical value, not a guess (Galaxy's own tool uses exactly this).
- `profile="20.01"` is functionally fine — expression tools landed in **19.05**, and Galaxy clamps with `if self.profile < 19.05: self.profile = 19.05`.

### Before trusting the green lint check

`tool_type="expression"` disables most of the linter suite: `lib/galaxy/tool_util/lint.py` skips any linter module whose `lint_tool_types` doesn't include the tool type, and for expression tools that leaves:

| runs | skipped |
|---|---|
| `general`, `inputs`, `containers` (`lint_tool_types=["*"]`) | `tests`, `citations`, `command`, `help`, `output`, `xml_order`, `xsd`, `required_files`, `datatypes` |

So lint here only exercised ID/name/profile/version. That makes issues **5** and **6** below reviewer-only — CI will never surface them.

---

## Blocking

**1. `parse_parameters.sh` is dead code.** Nothing references it (grep for `parse_parameters` matches only the XML). It implements the old shell column-splitting approach this tool replaces. Remove it.

**2. `weighting` silently parses wrong for every value except `"true"`.** The comparison `result['weighting'] == 'true'` is exact and case-sensitive:

| CSV cell | output |
|---|---|
| `true` | `true` |
| `TRUE` | `false` ← silent |
| `True` | `false` ← silent |
| `1` | `false` ← silent |

This matters concretely: the suite's own boolean convention is `truevalue="TRUE"` (`macros.xml:74`, `macros.xml:151`). A sweep table produced in the style Galaxy itself emits yields the **opposite** peak-detection setting with no error and no warning — silently different scientific results across a sweep. Normalise (`v.trim().toLowerCase()`) and reject unrecognised values rather than defaulting to `false`.

**3. CRLF and `", "`-style files fail with a misleading error.** The header is split on `,` but never trimmed, so cells keep `\r` (or a leading space), the lookup keys stop matching the clean names, and the *value* lookup then misses. Reproduced:

```
CRLF upload          → Parameter group_threshold does not contain a valid float value.
spaces after comma   → Parameter sigma_lower does not contain a valid float value.
```

Any user-authored CSV from Excel on Windows hits this, and the error points at a column that is fine. Strip `\r` and trim each header and value cell.

**4. `parseFloat` prefix-parses garbage, and `Infinity` becomes `null`.** `parseFloat` reads a leading number and ignores the rest, and `isNaN(parseFloat(x))` accepts it. Reproduced:

```
min_run = "3.1xyz"   → 3.1     (silently)
min_run = "1.2.3"    → 1.2     (silently)
min_run = "Infinity" → null    ← JSON.stringify(Infinity) === null
min_run = "1e400"    → null
```

Those `null`s propagate into downstream aplcms tools as missing parameters. A strict check is better than `parseFloat` here — and note `Number()` is *not* the fix, since `Number("") === 0` would silently accept empty cells. An explicit numeric regex on the trimmed cell (`/^-?\d+(\.\d+)?([eE][-+]?\d+)?$/`) rejects empty, `1.2.3`, `3.1xyz` and `Infinity` alike.

The common thread in 2-4 is the one failure mode a parameter-sweep helper must not have: plausible-but-wrong values with no error.

**5. Help text describes the deleted script, not this tool.** Line 162: *"It splits the CSV file into separate files for each column, excluding the header."* That is `parse_parameters.sh`'s behaviour, not this tool's (which emits one typed value per column from a single row). Also leftover scaffold: the `[ctrl+alt+t]` hint comment (104), the `TODO: auto-generated test case` comment (107), and `<citations></citations>` (184) where every sibling uses `<expand macro="citations"/>`.

**6. Zero of the five `throw`s are tested.** One happy-path test; all the validation logic is unverified. `expect_failure` is supported on `<test>`. Suggested cases, all confirmed to behave distinctly: wrong column count, wrong column name, non-numeric cell, header-only file, `TRUE` vs `true`.

**7. Error style diverges from every precedent.** This uses `throw new Error(...)`. The established convention is `return {'__error_message': '...'}`, which `handle_job.py` turns into a clean `Exception(result["__error_message"])`. It's used by `pick_value`, `map_param_value` and `regex_switch`; `throw` by none. A bare `throw` still fails the job, but the message goes through the JS worker's generic error path and is truncated to 500 chars — losing exactly the "which column, what value" detail that makes these errors actionable. Pairs with issues 3 and 4: the fix is not just better messages but the channel that delivers them intact.

---

## Should fix

**8. A multi-row CSV silently uses only row 1.** Reproduced: extra data rows are accepted and the second row is ignored. For a tool whose whole purpose is "one row of a sweep table", silently accepting a whole table and returning the first row is the wrong failure mode — a workflow that forgets to split would produce *N identical sweeps*. `metadata.data_lines` is set by `BaseCSV.set_meta` (`tabular.py:1478`), so `if (metadata.data_lines !== 1) throw ...` is a cheap, reliable guard.

**9. Five of nine column names don't match the tools that consume them.** Those names appear nowhere else in the repo (verified by grep):

| this tool | nearest downstream param (`macros.xml`) |
|---|---|
| `weighting` | `intensity_weighted` (:74) |
| `group_threshold` | `grouping_threshold` (:72) |
| `min_sd` / `max_sd` | `sd_cut_min` / `sd_cut_max` (:106,108) |
| `sigma_lower` / `sigma_higher` | `sigma_ratio_lim_min` / `sigma_ratio_lim_max` (:94,96) |
| `min_run`, `min_pres`, `mz_tol` | match ✅ |

The *mismatch* is a fact; the right-hand column is my inference about intent, since the sweep workflow isn't in this repo. Either align the names or document the mapping explicitly in the help — as written a workflow author has no way to confirm which output wires where.

**10. Question I can't answer from the repo: what unit is `mz_tol`?** Downstream, `mz_tol` is labelled `[ppm]` with default `10`, and its help says *"expressed as the fraction of the m/z value… multiplied by the m/z value, becomes the cutoff level"* — that existing help is itself self-contradictory (`macros.xml:66-69`). The test CSV here uses `0.01`. If the sweep table holds a *fraction* (1%) while the consumer expects *ppm*, every branch of the sweep is wrong by ~10⁶. Worth confirming, and worth fixing the upstream help contradiction either way.

**11. No range validation, though downstream declares ranges.** Reproduced: `min_pres = 5` is accepted despite downstream declaring `min="0.0" max="1.0"` (`macros.xml:60`); negative `min_run` is accepted; `min_sd=0.9, max_sd=0.1` (inverted pair) is accepted. Either validate against the same bounds or state that validation is deliberately deferred.

**12. Two independent sources of truth that can disagree.** The nine column names are written four times: `expectedColumnNames`, `floatParams`, the `return` object, and the `<outputs>` block. The script validates against `metadata.column_names` but then re-parses the file's own header text into `result`, so the metadata check is decorative and the two can diverge (metadata OK + differing text → error at *value*-lookup time, not at the header check). Since the full text is already in hand, I'd pick one source and collapse the schema into a single array of `{name, type, min?, max?}` records — that turns 3, 4, 8 and 11 into one code path.

---

## Minor

- `load_contents="64000"` is correct convention, but this file is header + one row (~200 bytes). With a fan-out of hundreds of sweep rows that's per-job overhead worth trimming.
- `name="row"` describes a whole CSV file, not a row; `argument="--row"` is inert here (`parse_command` hard-codes the command, so there is no command line). `name="row"` is the clearer spelling for a tool with no command.
- The `help="..."` attribute (89-90) contains a literal newline and a stray `]` inside the quoted header list — it will render oddly in the UI.
- Output assertions should use `value_json` — these outputs are JSON files, and `<output value_json="1.5"/>` is supported (real usage in `map_param_value`: `<output name="output_param_text" value_json="&quot;B&quot;" />`). The current `<has_text text="1.5"/>` is a substring match, so `1.55` would pass. Matters for a tool whose whole job is emitting numbers.
- Consistency with the 8 siblings: `profile="20.01"` vs `23.0`; locally-defined `@TOOL_VERSION@ 0.1.0` vs `macros.xml:2`'s `0.15.0`; `license="MIT"` set here only; inline `<help>` vs the suite's `@*_HELP@` tokens. The version decoupling is *defensible* — this tool doesn't wrap the R package — but deserves a comment, since one tool reading `0.1.0` next to siblings at `0.15.0` invites a bug report.
- `tool_type="expression"` sets `tool_type_local = True` (`__init__.py:3226`), so these jobs get the `local` dispatch class and the server needs a JS runtime + `cwl-utils`. Fine here (25.1 CI passes), but worth confirming against RECETOX's `job_conf.xml`.

---

## On generalising this (re: @bgruening's comment)

One technical constraint that's worth having on the record: **outputs must be statically declared** — `ExpressionTool.parse_outputs` raises `"Expression tools may not declare output datasets at this time."` for any output lacking `from_expression` (`__init__.py:3230-3233`), so outputs cannot be generated from the input's columns. That's the real reason this is aplcms-specific rather than generic.

Prior art: Galaxy core ships `tools/expression_tools/parse_values_from_file.xml` (id `param_value_from_file`, in the default panel at `tool_conf.xml.sample:62-63`), which reads `contents` the same way but coerces a whole file into **one** typed parameter. tools-iuc has 5 expression tools (`regex_switch`, `compose_text_param`, `pick_value`, `map_param_value`, `calculate_numeric_param`), none of which reads dataset contents into parameters. So there is no N-column variant to reuse.

The achievable middle ground: keep static outputs but drop the hard-wired aplcms column list, driving it from a `<repeat>` of column name + type (mirroring `expression_macros.xml`'s typed-output pattern), and fix the header parsing while there. My suggestion is to ship the specific version here and open the generic version as a separate issue with the `parse_outputs` constraint documented, rather than block this on an upstream API limitation.

**Bottom line:** the architecture is right and the happy path genuinely works in a real Galaxy. What blocks merge is not design but silent-wrong-value risk (2, 3, 4, 8), plus the dead script and the stale help text.

PR #728, opened 2026-01-29
AOP-tk wrappers for download, chemical identification + matching + normalization, PDF parsing, relationship finding

rdurnik:
AOP-tk wrappers for download, chemical identification + matching + normalization, PDF parsing, relationship finding


tools/aoptk/aoptk_chemical_normalization.xml
```diff
@@ -0,0 +1,56 @@
...
+<configfiles>
+<configfile name="normalize_chemicals">
+from aoptk.normalization.pubchem_api import PubChemAPI
+from aoptk.chemical import Chemical
+import os
+import pandas as pd
+
+chemicals_df = pd.read_csv("$input_file", sep="\t")
+chemicals_df["chemical"] = chemicals_df["chemical"].apply(
+    lambda x: PubChemAPI().normalize_chemical(Chemical(x))
```
rdurnik: PubChemAPI not yet in the 0.2.0 aoptk version... waiting for the new release.

mbodova (commented):
Hi, there seem to be several problems:
- test-data folder is not supposed to be specified in the tests, planemo test natively searches for files in that directory, e.g.:
```
<test>
    <param name="input_file" value="chemicals.tsv"/>
     ...
</test>
```
- After that some tests still fail, some of it might be just in my local environment, so I would start with removing 'test-data/' from paths in tests, adding missing output file and then seeing what fails on CI

These are the errors I encountered after fixing paths:
[tool_test_output.html](https://github.com/user-attachments/files/27437038/tool_test_output.html)

- For test in aoptk_finalize_output final_output.xml is missing
- In aoptk_normalize_chemicals and aoptk_find_relationship_chemicals LITELLM_CONFIG_FILE environment variable cannot be found
- In aoptk_download_publication - there are 2 types of error and I think they might be caused by this article 41480994 no longer being available in pmc
- In aoptk_publication_tracker tests fail due to difference in size, but also NCBI_API_KEY is not set - I don't know if it relates or not

hechth:
@arash77 this tool uses credentials to save and supply an openai api key and endpoint - how would I go about testing this, do you have any suggestions? I'd be really grateful!

arash77:
> @arash77 this tool uses credentials to save and supply an openai api key and endpoint - how would I go about testing this, do you have any suggestions? I'd be really grateful!

Unfortunately, there isn't any way to test it in the GitHub environment with real credentials for now, so you will have to either test it locally or simulate a failure.

hechth:
thanks @arash77 for the quick reply - then we will go with failure for now until we have some better options I think.

PR #791, opened 2026-07-15
aoptk: use column selects where it makes sense

### LINKED ISSUE #790: aoptk: use column selects where it makes sense
Many tools take a tabular file as input and require one or two columns to be used for matching or annotation etc. - the colnames are right now hardcoded.

It would be much easier if a user could select the column to choose for the process at hand (like the column identifying the chemical - the user can then choose chemical or heading, dependening on what is there etc.)

rdurnik:
Introduced ability to manually select a column for output summarization and normalization (in case of normalization using LLMs, also for the chemical list used for the normalization).


hechth:
@rdurnik you still need to bump the version of the wrapper (version suffix) - otherwise looks good!

hechth (approved):

